PR #4, opened 2017-02-26
make sure options are an object

shaharmor:


not-an-aardvark:
Thanks for the pull request.

Could you clarify what problem this is solving? Based on the [rule schema](https://github.com/not-an-aardvark/eslint-plugin-prettier/blob/3fd855dfb356c8616c19b51b70eb5fcb8fb90c9c/lib/rules/prettier.js#L15), options is already required to be an object.

shaharmor:
I don't know about the schema, but if using this in the `.eslintrc` file:
```js
{
    "rules": {
        "prettier/prettier": "error"
    }
}
```

the `context.options` is `[]`, which causes an error down the road in prettier because it expects the options to be an object.

when using custom options like this:
```js
{
    "rules": {
        "prettier/prettier": ["error", {"trailingComma": true, "singleQuote": true}]
    }
}
```

it works (the `context.options` is now `[{"trailingComma": true, "singleQuote": true}]`)

not-an-aardvark:
But if `context.options` is `[]`, then the call will just be `format(text, undefined)`. According to the [prettier docs](https://github.com/prettier/prettier#api) the argument is optional, so I would expect this to still work.

not-an-aardvark:
Oh, it looks like the issue is a bug in prettier: https://github.com/prettier/prettier/issues/801.

So this should be fixed the next time prettier does a release.

lydell:
Workaround: `"prettier/prettier": "error"` → `"prettier/prettier": ["error", {}]`.

PR #10, opened 2017-03-22
Running Tests

jasonLaster:
Adds a description for debugging unit tests with chrome devtools

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,3 +72,11 @@ To install `prettier`, use:
 ```bash
 npm install prettier --save-dev
 ```
+
+## Running Tests in the Browser
+
+The unit tests can be debugged with chrome devtools by adding a `debugger` statement in a test and running this command:
+
+```
+node --debug-brk  --inspect ./node_modules/mocha/bin/_mocha tests --recursive
+```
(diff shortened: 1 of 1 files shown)


not-an-aardvark:
Thanks for contributing!

What do you mean by adding a debugger statement to a test? The tests use ESLint's `RuleTester` utility, so the tests are just lists of cases and they don't individually have any statements.

jasonLaster:
I'll add an example. But feel free to reject if it isn't helpful. I added a debugger statement in the src code and was able to debug it to see what was going on. 

You could also add a debbuger statement in a test

jasonLaster:
I updated the language to be a little clearer. Sorry that it is still pretty verbose.

Here's a screenshot of what it looks like. Honestly, debugging node scripts with chrome devtools and soon firefox devtools is kinda my "hammer", but the docs are also in babel and other places and I always find it helpful when someone points out that it's possible.

screen shot 2017-03-22 at 10 34 41 pm

zertosh:
Thanks for the PR @jasonLaster! While I'm sure some people would find this information helpful, having it as part of the plugin docs feels out of place. I think the node docs on this topic are just fine https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js

PR #15, opened 2017-05-01
New: allow using forks of `prettier`

josephfrazier:
This makes it possible for projects to specify the name of a `prettier`
fork to be used instead of the original `prettier`. The fork name is
passed as an additional option after the `prettier` configuration. For
example:

```json
{
    "rules": {
        "prettier/prettier": ["error", {"trailingComma": "es5", "singleQuote": true}, "prettier-with-tabs"]
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ If you'd like to use a fork of `prettier`, pass its name as an additional option
 ```json
 {
     "rules": {
-        "prettier/prettier": ["error", {"trailingComma": "es5", "singleQuote": true}, "prettier-with-tabs"]
+        "prettier/prettier": ["error", {"trailingComma": "es5", "singleQuote": true}, "prettier-miscellaneous"]
     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -57,6 +57,16 @@ You can also pass [`prettier` configuration](https://github.com/prettier/prettie
 }
 ```
 
+If you'd like to use a fork of `prettier`, pass its name as an additional option:
+
+```json
+{
+    "rules": {
+        "prettier/prettier": ["error", {"trailingComma": "es5", "singleQuote": true}, 'prettier-with-tabs']
+    }
+}
+```
+
 The rule will report an error if your code does not match `prettier` style. The rule is autofixable -- if you run `eslint` with the `--fix` flag, your code will be formatted according to `prettier` style.
 
 ---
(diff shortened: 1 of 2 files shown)


not-an-aardvark:
Thanks for the pull request!

Hmm, I'm not sure about this, for a few reasons:

* My first impression is that this functionality is out-of-scope for this project. This is intended to be a plugin for prettier, not arbitrary libraries that expose a similar API. I'm not sure there is enough demand for this to be worthwhile.
* Implementation-wise, passing the name of a module in the config seems like a leaky abstraction because `require()` will be called from `eslint-plugin-prettier`, not the module that created the config. This would result in unexpected changes in behavior if a module with the same name is added to `eslint-plugin-prettier`'s dependency tree.

josephfrazier:
Thanks for the feedback!

> My first impression is that this functionality is out-of-scope for this project. This is intended to be a plugin for prettier, not arbitrary libraries that expose a similar API. I'm not sure there is enough demand for this to be worthwhile.

That's fair. I didn't intend this to be used for "arbitrary libraries" with the same API, but it could of course be used in that way. I'd like to note that there is at least one fork, [prettier-miscellaneous](https://github.com/arijs/prettier-miscellaneous/tree/dbdb984fbdb18d6efb6307dfcf4a355f1dd945ad), that is mentioned in the official documentation under [Related Projects](https://github.com/prettier/prettier/tree/515a565a66b458b8a3f40d0878ce856e4acf5f9f#related-projects) (it's the one I'd like to try using with eslint via this plugin). I also updated the docs in this PR to reflect the new name of that project, rather than `prettier-with-tabs`.

> Implementation-wise, passing the name of a module in the config seems like a leaky abstraction because require() will be called from eslint-plugin-prettier, not the module that created the config. This would result in unexpected changes in behavior if a module with the same name is added to eslint-plugin-prettier's dependency tree.

Oh, good catch. I've updated the PR to resolve the module relative to the current working directory, rather than the eslint-plugin-prettier directory.

---

I hope you'll still consider this. Let me know what you think!

not-an-aardvark:
Unfortunately, I'm going to close this -- this plugin is only designed to work with `prettier`, and I think passing a module name as a string could lead to confusing behavior for users. For example, resolving the fork module from the cwd solves the issue described in https://github.com/not-an-aardvark/eslint-plugin-prettier/pull/15#issuecomment-298499719, but then the plugin would sometimes not work if it was run with a CWD outside the project, or if a config file was in a different directory than the file being linted.

Thanks again for being willing to contribute! Feel free to fork this module if you want something that works with `prettier-with-tabs`.

josephfrazier:
That makes sense. Thanks for the consideration!

Issue #17, opened 2017-05-10
line-by-line error reporting


vjeux:
Hey @not-an-aardvark, thanks a ton for working on this plugin, this turned out to be the way most people are integrating prettier inside of their stack :)

We've been working on a different implementation at Facebook. At first we started it to support annotating files with `@format` (like `@flow`) and only running the lint rule on them, for a migration path.

In the process, we added a super nice feature where instead of reporting a single eslint message at the top of the file, we diff the file and the prettier output and give suggestions for each line that needs to be changed. When you have an editor integration, it shows it inline like any other lint rule, which is really nice.

https://github.com/zertosh/eslint-plugin-prettify

I was wondering if we could figure out a way to get those two projects to converge, so that people using prettier have the best possible integration :)

Thanks!

zertosh:
This is a huge ask, but maybe @not-an-aardvark is open to moving `eslint-plugin-prettier` to the Prettier org - then I could deprecate `eslint-plugin-prettify` and move over the code to `eslint-plugin-prettier`.

not-an-aardvark:
Hi, thanks for the issue.

> In the process, we added a super nice feature where instead of reporting a single eslint message at the top of the file, we diff the file and the prettier output and give suggestions for each line that needs to be changed. When you have an editor integration, it shows it inline like any other lint rule, which is really nice.

Seems useful! This is similar to what I had in mind for https://github.com/not-an-aardvark/eslint-plugin-prettier/issues/7, but I never ended up implementing it.

> This is a huge ask, but maybe @not-an-aardvark is open to moving `eslint-plugin-prettier` to the Prettier org - then I could deprecate `eslint-plugin-prettify` and move over the code to `eslint-plugin-prettier`.

Sure, that sounds good to me. I haven't had much time to maintain this recently, so if other people want to make improvements/help maintain the plugin, that would be a win-win situation.

not-an-aardvark:
I just accepted the invitation to the prettier org -- should I go ahead and transfer the repo?

vjeux:
Yup, should be good to go!

not-an-aardvark:
Done.

vjeux:
Woot woot!

vjeux:
@zertosh is splitting the work he has done on his repo into distinct pull requests that he will send here. If you are happy with them, then we can start merge them :)

It has a compatible API, so it shouldn't disrupt existing users, they are just going to get a better experience out of the box and will be able to use more options :)

PR #21, opened 2017-05-15
Merge with eslint-plugin-prettify

zertosh:
As discussed in #17, [eslint-plugin-prettify](https://github.com/zertosh/eslint-plugin-prettify) will merge into eslint-plugin-prettier. eslint-plugin-prettify diffs the original code against the prettier output, and reports fine grained issues, instead of invalidating the entire file.

This PR ports the code at https://github.com/zertosh/eslint-plugin-prettify/commit/2fb5dfbb52a2577180bf8fcccb83b3fa7aa4d69c, and adjusts it to conform to @not-an-aardvark's pre-existing code style. A few notes:

* The new rule is backwards compatible with the previous one. The option can be an object that is passed to Prettier. In addition to that, it can now be the string `"fb"` - which sets preferred "Facebook style" (see the README).
* A new second option can be passed to the rule, which sets a "pragma" requirement for linting. If it's omitted, then all files are linted. If it's included, then only files with that pragma are linted. eslint-plugin-prettify automatically set a pragma of `@format` if you used the magic `"fb"` config. In this port, that's no longer the case (as requested by @kittens).
* eslint-plugin-prettify had a special entry `no-styles`, which added a preprocessor that stripped style rules from the final result. That's not included in this PR, it'll be added in a followup PR.


.eslintrc.js
```diff
@@ -1,17 +1,11 @@
...
-const PACKAGE_NAME = require('./package').name;
-const SYMLINK_LOCATION = path.join(__dirname, 'node_modules', PACKAGE_NAME);
-
-// Symlink node_modules/{package name} to this directory so that ESLint resolves this plugin name correctly.
-if (!fs.existsSync(SYMLINK_LOCATION)) {
-  fs.symlinkSync(__dirname, SYMLINK_LOCATION);
-}
+// Register ourselves as a plugin to avoid `node_modules` trickery.
+const Plugins = require('eslint/lib/config/plugins');
+Plugins.define('prettier', require('.'));
```
not-an-aardvark: I think this approach is going to break after https://github.com/eslint/eslint/pull/8465 is merged, which is likely to happen in ESLint 4. (The change makes `Plugins` an ES6 class rather than a global mutable object, to prevent unrelated modules from conflicting with each other if they both happen to be using eslint in the same process.)

README.md
```diff
@@ -1,78 +1,118 @@
...
+      "jsxBracketSameLine": true,
+      "parser": "flow"
+    }]
+    ```
 
-This plugin works best if you disable all other ESLint rules relating to code formatting, and only enable rules that detect patterns in the AST. (If another active ESLint rule disagrees with `prettier` about how code should be formatted, it will be impossible to avoid lint errors.) You can use [eslint-config-prettier](https://github.com/lydell/eslint-config-prettier) to disable all formatting-related ESLint rules. If your desired formatting does not match the `prettier` output, you should use a different tool such as [prettier-eslint](https://github.com/kentcdodds/prettier-eslint) instead.
+* The second option:
 
-## Migrating to 2.0.0
+  - A string with a pragma that triggers this rule. By default, this rule applies to all files. However, if you set a pragma (this option), only files with that pragma in the heading docblock will be checked. All pragmas must start with `@`. Example:
```
not-an-aardvark: To me, it seems like it would be better to get this functionality by using normal `eslint` directive comments, e.g.

```js
/* eslint prettier: ["error", "fb"] */
```

Is there a reason to have a special pragma for the rule?

eslint-plugin-prettier.js
```diff
@@ -0,0 +1,306 @@
...
+  for (let i = 0; i < str.length; i++) {
+    switch (str[i]) {
+      case ' ':
+        ret += '\u00B7'; // Middle Dot
+        break;
+      case '\n':
+        ret += '\u23ce'; // Return Symbol
+        break;
+      case '\t':
+        ret += '\u21b9'; // Left Arrow To Bar Over Right Arrow To Bar
```
not-an-aardvark: Nitpick: Maybe this would be more readable if you used the symbols themselves in the code rather than their escape codes. For example:

```js
switch (str[i]) {
  case ' ':
    ret += '·'; // Middle Dot, \u00b7
    break;
  case '\n':
    ret += '⏎'; // Return Symbol, \u23ce
  case '\t':
    ret += '↹'; // Left Arrow To Bar Over Right Arrow To Bar, \u21b9
}
```

That way, if someone sees a symbol in the output, it's easier to tell where in the code the symbol is coming from.

eslint-plugin-prettier.js
```diff
@@ -0,0 +1,306 @@
...
+  //           \-> DELETE |
+  //                      \-> INSERT -> EQUAL
+  // Instead of reporting issues at each INSERT or DELETE, certain sequences
+  // are batched together and are reported as a friendlier "replace" operation:
+  // - A DELETE immediately followed by an INSERT.
+  // - Any number of INSERTs and DELETEs where the joining EQUAL of one's end
+  // and another's beginning does not have line endings (i.e. issues that occur
+  // on contiguous lines).
+
+  const source = context.getSource();
```
not-an-aardvark: I think `context.getSource()` is deprecated -- there doesn't seem to be any documentation for it, and it's in the list of deprecated passthroughs [here](https://github.com/eslint/eslint/blob/0ef09ea071f7df0cfb3f76ddea9ad746cb619912/lib/rule-context.js#L24-L41). Can you replace it with `context.getSourceCode().text` instead?

eslint-plugin-prettier.js
```diff
@@ -0,0 +1,306 @@
...
+/**
+ * Reports an "Insert ..." issue where text must be inserted.
+ * @param {RuleContext} context - The ESLint rule context.
+ * @param {number} offset - The source offset where to insert text.
+ * @param {string} text - The text to be inserted.
+ * @returns {void}
+ */
+function reportInsert(context, offset, text) {
+  const pos = getLocFromIndex(context, offset);
+  const range = [null, offset];
```
not-an-aardvark: I'm not sure `null` is documented as a valid value in a fix range. Can you replace this with `[offset, offset]` or something like that instead?

(Admittedly, the design of `insertTextAfterRange` does make the first value in the range useless, but I don't want this to break if ESLint starts validating ranges later on by asserting that both values are numbers.)

eslint-plugin-prettier.js
```diff
@@ -0,0 +1,306 @@
...
+// Preferred Facebook style.
+const FB_PRETTIER_OPTIONS = {
+  singleQuote: true,
+  trailingComma: 'all',
+  bracketSpacing: false,
+  jsxBracketSameLine: true,
+  parser: 'flow'
+};
+
+const LINE_ENDING_RE = /\r\n|[\r\n\u2028\u2029]/g;
```
not-an-aardvark: Can this be a non-global regex? It seems like it's only used as `LINE_ENDING_RE.test(something)`.

eslint-plugin-prettier.js
```diff
@@ -0,0 +1,306 @@
...
+        : context.options[0];
+
+      const pragma = context.options[1]
+        ? context.options[1].slice(1) // Remove leading @
+        : null;
+
+      if (pragma) {
+        // The pragma is only valid if it is found in a block comment at the
+        // very start of the file.
+        const firstComment = context.getAllComments()[0];
```
not-an-aardvark: I think `context.getAllComments()` is also deprecated. Can you use `context.getSourceCode().getAllComments()` instead?

Also, starting in ESLint 4, shebang comments will be included in the result of `sourceCode.getAllComments` (see the [migration guide](http://eslint.org/docs/4.0.0/user-guide/migrating-to-4.0.0#shebangs)). Shebang comments need to be placed at the start of a file, so maybe it would be better to filter them here.

test/prettier.js
```diff
@@ -0,0 +1,92 @@
...
+  const filename = path.join(__dirname, 'invalid', name + '.txt');
+  const src = fs.readFileSync(filename, 'utf8');
+  const sections = src
+    .split(/^[A-Z]+:\n/m)
+    .map(x => x.replace(/(?=\n)\n$/, ''));
+  const item = {
+    code: sections[1],
+    output: sections[2],
+    options: eval(sections[3]), // eslint-disable-line no-eval
+    errors: eval(sections[4]) // eslint-disable-line no-eval
```
not-an-aardvark: Would it be possible to use `JSON.parse` here instead? It seems like there's no reason to execute array literals as code.

test/prettier.js
```diff
@@ -0,0 +1,92 @@
...
+    { code: '""\n', options: [null, '@format'] },
+    // Facebook style uses single quotes.
+    { code: `'';\n`, options: ['fb'] },
+    // Facebook style but missing pragma.
+    { code: `"";\n`, options: ['fb', '@format'] },
+    // Facebook style with pragma.
+    { code: `/** @format */\n'';\n`, options: ['fb', '@format'] }
+  ],
+  invalid: [
+    '01',
```
not-an-aardvark: This isn't a strong objection, but I have doubts about whether putting the invalid cases in separate files actually improves debuggability. If a single test case is failing, it seems like it would be easier to search for the corresponding code within a single file rather than opening/closing a bunch of test files to try to find the right one.

test/prettier.js
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env node
+/* eslint-disable node/shebang */
```
not-an-aardvark: Why is this file executable? `RuleTester` uses mocha internally, so it seems like this would throw an error if it was just run with `node`.

eslint-plugin-prettier.js
```diff
@@ -0,0 +1,306 @@
...
+        }
+      }
+
+      return {
+        Program() {
+          if (!prettier) {
+            // Prettier is expensive to load, so only load it if needed.
+            prettier = require('prettier');
+          }
+          const source = context.getSource();
```
not-an-aardvark: I think `context.getSource()` is deprecated -- there doesn't seem to be any documentation for it, and it's in the list of deprecated passthroughs [here](https://github.com/eslint/eslint/blob/0ef09ea071f7df0cfb3f76ddea9ad746cb619912/lib/rule-context.js#L24-L41). Can you replace it with `context.getSourceCode().text` instead?

eslint-plugin-prettier.js
zertosh: 👍

eslint-plugin-prettier.js
zertosh: 👍

test/prettier.js
zertosh: It works actually, you just don't get pretty output if there's an error. But I'll remove it since that's an odd use case.

test/prettier.js
zertosh: Ah, it's actually more about easily looking at a fixture file and easily converting line & column in the test vs the message. It's also very convenient to be able to quickly focus on a few tests by commenting out.

test/prettier.js
zertosh: JSON formatting is too restrictive: Quoting keys, and using double quotes for strings. RuleTester's output is not JSON, so copy/pasting expected output would've required re-formatting. If it's all the same to you, I'd like to keep the `eval` until it becomes a problem.

.eslintrc.js
not-an-aardvark: There's an accepted issue for it at https://github.com/eslint/eslint/issues/6237, but when I looked into it a little while ago it seemed like it could introduce problems even when absolute paths are used. Feel free to comment on that thread if you have any ideas for how to handle that, though.

Hmm, I hadn't realized modifying `node_modules` breaks yarn. Maybe we should keep `Plugins` somehow until we have a better way to refer to plugins by path.

zertosh: > Hmm, I hadn't realized modifying node_modules breaks yarn

It causes `yarn check` to fail. Also, it'll break a new yarn feature I'll be upstreaming soon: "watchman-based check". Basically, you get instant (sub ~100ms) full `node_modules` verification, which lets use put `yarn` in front of every js script w/o having to worry about stale modules.

test/prettier.js
not-an-aardvark: Huh, TIL.

zertosh:
I enabled "Rebase and merge" to preserve the commits on this PR. If it all looks good, then I can merge and publish.

not-an-aardvark:
By the way, I've been using [this script](https://github.com/not-an-aardvark/eslint-config/blob/c4518a7a61eb25d23f2d89aa1ccccaf3b9f5027c/build/release.js) to autogenerate changelogs and git tags, in case we want to keep using it. (It requires a few additional devDependencies that `eslint-plugin-prettier` currently doesn't have. I'd been running the script locally with a relative path outside of the  `eslint-plugin-prettier` repo, but if we want to keep using it, it might be better to put it in the repo.)

zertosh:
neat script @not-an-aardvark! I just used it and pushed the results :) I'll check it into the repo & add some release instructions. Mind doing an npm release and/or adding me as a collaborator?

not-an-aardvark:
Will do. Are you `zertosh` on npm?

zertosh:
@not-an-aardvark, yup! Thanks!

not-an-aardvark:
Thanks for confirming. I'm on mobile right now but I'll add you sometime in the next few hours.

not-an-aardvark:
Added you to the npm package.

Issue #19, opened 2017-05-11
inter] Error running ESLint Error: require(...).getLocationFromRangeIndex is not a function(…)


christianalfoni:
Hi there!

I have installed this plugin as instructed on the repo. I am using latest version of Atom with latest version of linter-eslint, and also latest eslint installed to my project. Also latest prettier.

For some reason when I save I get the error in the title of this issue. Not sure where to start here :) The error goes away when I remove the prettier rule from the rules map.

not-an-aardvark:
Thanks for the report. This also occurred in https://github.com/prettier/eslint-plugin-prettier/issues/11, although I don't think we ever figured out why. It seems like restarting Atom fixed the issue in https://github.com/prettier/eslint-plugin-prettier/issues/11.

christianalfoni:
Hi and thanks for the answer :-)

What I notice is that this only happens when I have React code in my file. I have tried to restart and it did not help. Hm hm...

This is my config file:

```js
module.exports = {
  "parserOptions": {
    "ecmaVersion": 2016,
    "sourceType": "module",
    "ecmaFeatures": {
      "jsx": true
    }
  },
  "env": {
    "es6": true,
    "node": true
  },
  plugins: [
      "prettier",
      "react"
  ],
  extends: [
    "plugin:react/recommended",
    "prettier",
    "prettier/react"
  ],
  rules: {
    "prettier/prettier": "error"
  }
};
```

christianalfoni:
Okay, I figured it out:

It was my global eslint messing it up. Made sure the "Use global eslint" config at the bottom of the linter-eslint plugin for Atom was turned off and installed eslint and eslint-plugin-react again to my local project.

iameli:
I got the same error in VS Code, with a similar workaround -- I had `eslint-plugin-prettier` installed globally but not locally.

Mostly throwing this here 'cause it pops up first on Google for the error.

Issue #11, opened 2017-03-23
[Linter] Error running ESLint Error: require(...).getLocationFromRangeIndex is not a function(…)


nwaywood:
ESLint is failing to run for me with `eslint-plugin-prettier` rules activated (i.e. with `"rules": { "prettier/prettier": "error" }` set). 

This is the stacktrace from atom:

![image](https://cloud.githubusercontent.com/assets/3773641/24233265/3a20a572-0fe5-11e7-8c9b-321fe075744e.png)

As it says in the stacktrace, the error comes from [this](https://github.com/not-an-aardvark/eslint-plugin-prettier/blob/master/lib/rules/prettier.js) file on line 30 where it calls `require('eslint/lib/ast-utils').getLocationFromRangeIndex(...)`.

Upon inspection of [eslint/lib/ast-utils](https://github.com/eslint/eslint/blob/master/lib/ast-utils.js) I see that the function no longer exists. The function was removed in [this](https://github.com/eslint/eslint/commit/dad20ad940eedcc3f2e3f72b3f182dda5a8e700c#diff-6198b73cda8ab9ac49ed872ead7fb268) commit.

not-an-aardvark:
Does this still occur with `eslint-plugin-prettier@2.0.1`?

nwaywood:
Yes. Here are the devDeps from my `package.json`

![image](https://cloud.githubusercontent.com/assets/3773641/24233444/5e547cce-0fe6-11e7-91dd-92042d9a5884.png)

nwaywood:
Not sure if this is important/relevant but this error only happens when the file does not confirm to the prettier format. If a file is already formatted by prettier than the linter doesn't log this error

not-an-aardvark:
Hmm, that's odd. You're right that the function from `eslint/lib/ast-utils` no longer exists, but the ternary expression [here](https://github.com/not-an-aardvark/eslint-plugin-prettier/blob/master/lib/rules/prettier.js#L28-L33) is intended to check for the existence of `sourceCode.getLocFromIndex`, which was added in the same commit that the other function was removed.

What version of eslint do you currently have installed? (Your package.json just shows `^3.14.1`, but I'm wondering if there's a particular eslint version where this breaks. It seems to work fine for me with the latest version of eslint.)

nwaywood:
[...]

PR #24, opened 2017-05-18
Fix: Support ESLint <3.11.0

zertosh:
Converting from a range/index to a location varies between ESLint versions. This PR covers versions all the way back to ESLint 3.9.0, and creates a test matrix of the various versions.

Context: Unfortunately, we have a project with ESLint 3.9.0 that I can't update right now without making a mess.

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -45,15 +45,37 @@ let prettier;
  * @returns {Object} An object containing numeric `line` and `column` keys.
  */
 function getLocFromIndex(context, index) {
-  // If `sourceCode.getLocFromIndex` is available from ESLint, use it.
-  // Otherwise, use the private version from eslint/lib/ast-utils.
-  // `sourceCode.getLocFromIndex` was added in ESLint 3.16.0.
+  // If `sourceCode.getLocFromIndex` is available from ESLint, use it - added
+  // in ESLint 3.16.0.
   const sourceCode = context.getSourceCode();
   if (typeof sourceCode.getLocFromIndex === 'function') {
     return sourceCode.getLocFromIndex(index);
   }
-  const astUtils = require('eslint/lib/ast-utils');
-  return astUtils.getLocationFromRangeIndex(sourceCode, index);
+  const text = sourceCode.getText();
+  if (typeof index !== 'number') {
+    throw new TypeError('Expected `index` to be a number.');
+  }
+  if (index < 0 || index > text.length) {
+    throw new RangeError('Index out of range.');
+  }
+  // Loosely based on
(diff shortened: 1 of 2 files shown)


eslint-plugin-prettier.js
```diff
@@ -45,15 +45,45 @@ let prettier;
...
+  let offset = 0;
+  for (let i = 0; i < lines.length; i++) {
+    if (index >= offset && index <= offset + lines[i].length) {
+      return {
+        line: i + 1,
+        column: index - offset
+      };
+    }
+    // Add 1 for the line ending character
+    offset = offset + lines[i].length + 1;
```
not-an-aardvark: I think this will be incorrect if a file contains CRLF newlines, because the line ending will have a length of 2.

IIRC, there isn't enough information in the `lines` array to compute this accurately because it doesn't store which linebreaks were used. `astUtils.getLocationFromRangeIndex` was implemented by [splitting the text into lines again](https://github.com/eslint/eslint/blob/18a519fa2971e54a639e2b3320aeecd9a31b390a/lib/ast-utils.js#L408-L438) and including linebreaks in the result.

eslint-plugin-prettier.js
```diff
@@ -45,15 +45,45 @@ let prettier;
...
-  // `sourceCode.getLocFromIndex` was added in ESLint 3.16.0.
+  // If `sourceCode.getLocFromIndex` is available from ESLint, use it - added
+  // in ESLint 3.16.0.
   const sourceCode = context.getSourceCode();
   if (typeof sourceCode.getLocFromIndex === 'function') {
     return sourceCode.getLocFromIndex(index);
   }
+  // Otherwise, try to use the private version from eslint/lib/ast-utils -
+  // added in ESLint 3.11.0.
   const astUtils = require('eslint/lib/ast-utils');
```
not-an-aardvark: I haven't figured out the root cause yet, but I think using `require('eslint')` might be the source of #11 and #19. That could happen if the module found by running `require('eslint')` is not the same as the eslint module that is actually running the rule, which could result in neither `sourceCode.getLocFromIndex` nor `astUtils.getLocationFromRangeIndex` existing even when both modules are above 3.11.0.

That said, I guess this isn't a problem anymore because the fallback will get used now if neither function exists.

eslint-plugin-prettier.js
zertosh: Ah so true - can you tell I don't interact with windows much? 😛 I circle back to this tmw.

zertosh:
@not-an-aardvark, I removed the use of `eslint/lib/ast-utils`, and went with a CRLF-friendly fallback implementation of `getLocFromIndex` when it's not available.

PR #38, opened 2017-07-15
[WIP] Implement prettier-eslint integration

azz:
A bit of a work in progress/proof of concept.

image

Why not directly integrate `prettier-eslint` into this plugin? Currently `prettier-eslint` and this plugin are contradictory, but they could be made to work together.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ Then, in your `.eslintrc`:
     "prettier/prettier": ["error", {"singleQuote": true, "parser": "flow"}]
     ```
 
+  - Set `"prettier-eslint": true"` to use [`prettier-eslint`](https://github.com/prettier/prettier-eslint) instead of prettier.
+
+    ```json
+    "prettier/prettier": ["error", {"prettier-eslint": true}]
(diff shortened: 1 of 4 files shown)


not-an-aardvark:
Thanks for the PR. Could you clarify what problem this fix is trying to solve?

My initial thoughts are that I'm not sure this is a good idea, because it violates the separation of responsibilities between ESLint rules. (By design, all ESLint rules can be configured independently from each other, and rules can't read the configs of other rules.) Looking around the filesystem to guess the config can work in some cases, but it's not a good solution in general, because:

* it won't work for configs passed on the command line
* it won't work for text that doesn't have a filename (e.g. text piped from stdin, or when an integration is using the Node.js API).

azz:
Yeah you're probably right.

The problem I am trying to solve is the incompatibility between `prettier-eslint` and this plugin, providing errors when the output of `eslint --fix` after `prettier` do not match the source.

It was more a question of _could_ it be done, rather than _should_ it.

Issue #39, opened 2017-07-31
Expose report-related APIs in another package


ikatyang:
Currently, all the APIs are not exposed, so that it can't be reused by other linter-plugin, e.g. [tslint-plugin-prettier](https://github.com/ikatyang/tslint-plugin-prettier/tree/master/src/utils) (currently just copy-paste implementations from this repo).

It'd be better to expose report-related APIs in another package (since this repo has peerDeps for eslint), so that it can be reused in different linter plugin to have consistent behavior.

Suggested API

```ts
export declare function showInvisibles(str: string): string;
export declare function reportDifferences(source: string, formatted: string, reporters: Reporters);

export interface Reporters {
  reportInsert: (offset: number, text: string, message: string) => void;
  reportDelete: (offset: number, text: string, message: string) => void;
  reportReplace: (offset: number, deleteText: string, insertText: string, message: string) => void;
}
```

not-an-aardvark:
I wouldn't mind doing this in theory, but it feels like the helper functions are sufficiently small and eslint-specific that it wouldn't be worthwhile. For example, all of the helpers use `getLocFromIndex`, `context`, and `fixer`, which are ESLint APIs and (as far as I know) have no direct equivalent in tslint.

I don't think your proposed interface would work, because something like `reportInsert` would need to have some way of actually reporting an error, and that can't happen unless they either return a value or accept a report function.

ikatyang:
The `context` is not a necessary argument for reporters, for example if there is the suggested API:

```js
return {
  Program() {
    if (!prettier) {
      // Prettier is expensive to load, so only load it if needed.
      prettier = require('prettier');
    }
    const prettierSource = prettier.format(source, prettierOptions);
    if (source !== prettierSource) {
      suggestedAPI.reportDifferences(source, prettierSource, {
        reportInsert: (...args) => reportInsert(context, ...args),
        reportDelete: (...args) => reportDelete(context, ...args),
        reportReplace: (...args) => reportReplace(context, ...args),
      });
    }
  }
};
```

We can call `context` as a contextual value so that it wouldn't have to pass in reporters directly.

For TSLint vs ESLint API, see [this](https://github.com/ikatyang/tslint-plugin-prettier/blob/master/src/utils/report-replace.ts).

If you still think my proposed interface wouldn't work, I can send a PR here to show what it'd be in my proposed.

ikatyang:
See #40 for my proposal

[...]

PR #40, opened 2017-07-31
[DO NOT MERGE] feat: expose reporter api

ikatyang:
demo #39

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -9,8 +9,8 @@
 //  Requirements
 // ------------------------------------------------------------------------------
 
-const diff = require('fast-diff');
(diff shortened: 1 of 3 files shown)


zertosh:
I like the idea, but it seems like you're pulling out stuff into "utils" for the sake of pulling them out.

ikatyang:
This PR is just for demo, since the exposed API should in another package due to peerDeps. 😅

not-an-aardvark:
We could always remove the `eslint` peerDependency if we're exposing an API that works without eslint.

ikatyang:
OK, I'll reopen and modify this PR later.

ikatyang:
I somehow can't reopen this PR, so I create a new one #41.

PR #41, opened 2017-08-01
feat: expose reporter api

### LINKED ISSUE #39: Expose report-related APIs in another package

ikatyang:
Fixes #39

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
     'plugin:eslint-plugin/recommended',
     'prettier'
   ],
+  env: { mocha: true },
   root: true,
   rules: {
     'prettier/prettier': ['error', { singleQuote: true }],
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -129,13 +129,13 @@ function generateDifferences(source, prettierSource) {
   // and another's beginning does not have line endings (i.e. issues that occur
   // on contiguous lines).
 
-  const diffResults = diff(source, prettierSource);
-  const returnResults = [];
+  const results = diff(source, prettierSource);
+  const differences = [];
 
   const batch = [];
   let offset = 0; // NOTE: INSERT never advances the offset.
-  while (diffResults.length) {
-    const result = diffResults.shift();
+  while (results.length) {
+    const result = results.shift();
     const op = result[0];
     const text = result[1];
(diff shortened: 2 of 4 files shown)


zertosh:
Ok, so I had a bit of time to think about this and I'm not sure this is a great approach. Besides, the lack of docs and testing – to make sure this doesn't change w/o breaking you – an API that takes reporter functions creates a larger surface area than necessary.

What is now `reportDifferences`, should really be a `generateDifferences` that returns an array of insert/delete/replace ops, then you call your own reporter on them.

I'll put something together later today since it turns out I also kinda need this for a clang-format linter I'm working on.

ikatyang:
I think this package can be separated into three part (`eslint-plugin-prettier`, `generate-differences `and `show-invisible`), all of them are useful and reusable, but it seems you don't want to separate them, so I just refactor them to have minimum changes for exposing API.

maybe we can have something like this? 

```ts
function generateDifferences(source: string, target: string): Result[];
interface Result {
  operation: Operation;
  offset: number;
  insertText: string;
  deleteText: string;
}
enum Operation {
  Insert,
  Delete,
  Replace,
}
```

ikatyang:
@zertosh @not-an-aardvark, is this PR okay or what should I do?

zertosh:
I don't have anything constructive to add.

ikatyang:
Can we get this merged and publish? Thanks. 😅

not-an-aardvark:
Oops, I forgot to publish it. I'll do that now.

not-an-aardvark:
Published as v2.2.0.

Issue #46, opened 2017-08-29
Support prettier v1.6.0 config


jaydenseric:
[Prettier v1.6.0](https://github.com/prettier/prettier/releases/tag/1.6.0) is out, and it impliments consmiconfig for options.

This plugin currently requires prettier config to be set within ESLint config. It would be awesome to deprecate this and let prettier derive the config itself.

not-an-aardvark:
Seems reasonable to me. It looks like `prettier` exposes a [`resolveConfig`](https://github.com/prettier/prettier/blob/987b931215b7e2f23df037dfa98344f2993949ad/README.md#prettierresolveconfigfilepath--options) function for getting a config, so I think it should be pretty simple to let prettier handle config resolution.

not-an-aardvark:
Actually, I think this is blocked by https://github.com/prettier/prettier/pull/2722.

goldo:
@not-an-aardvark since it has been merged, I guess we have to wait for the next prettier release ?

not-an-aardvark:
Yes, that's correct.

azz:
Tracking for 1.6.2 which will include .sync() https://github.com/prettier/prettier/issues/2804

not-an-aardvark:
Support for prettier config files has been published in `eslint-plugin-prettier@2.3.0`. Thanks to @kombucha for implementing this.

PR #55, opened 2017-09-17
Fix: Support prettier v1.6.0 config (#46)

kombucha:
I'm not sure if anybody was working on this issue so I made this PR !
Tests are passing, and I tested it on a small example project.

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,7 @@ Then, in your `.eslintrc.json`:
 ## Options
 
 * The first option:
-
-  - Objects are passed directly to Prettier as [options](https://github.com/prettier/prettier#api). Example:
+  - Objects are passed directly to Prettier as [options](https://github.com/prettier/prettier#options). Example:
     
     ```json
     "prettier/prettier": ["error", {"singleQuote": true, "parser": "flow"}]
@@ -78,6 +77,7 @@ Then, in your `.eslintrc.json`:
       "parser": "flow"
     }]
     ```
+  NB: This option will merge and override any config set with `.prettierrc` files (for Prettier < 1.7.0, [config files are ignored](https://github.com/prettier/eslint-plugin-prettier/issues/46))
 
 * The second option:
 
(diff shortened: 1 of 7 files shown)


eslint-plugin-prettier.js
```diff
@@ -342,12 +337,31 @@ module.exports = {
           }
         }
 
+        if (prettier) {
+          prettier.clearConfigCache();
```
kombucha: Not too sure about this but it proved necessary when using vscode, otherwise changes to the `.prettierrc` were not getting picked up.

README.md
```diff
@@ -55,7 +55,7 @@ Then, in your `.eslintrc.json`:
 
 * The first option:
 
-  - Objects are passed directly to Prettier as [options](https://github.com/prettier/prettier#api). Example:
+  - Objects are passed directly to Prettier as [options](https://github.com/prettier/prettier#api). This plugin will merge and override any config set with a `.prettierrc` file (only for prettier >= 0.17). Example:
```
lydell: - Here's a better link for options: https://github.com/prettier/prettier/#options
- You could change "prettier" to "Prettier" for consistency :)
- Typo: `0.17` → `1.7.0`
- I think it is clearer to say `[Prettier configuration file](https://github.com/prettier/prettier/#configuration-file)` instead of `` `.prettierrc` file ``. Not sure if we need to mention the version at all.
- The sentence about config file support applies both to the object option and the `"fb"` option. I suggest moving it outside of all the bullet lists to the very start of the "Options" section.

kombucha: I fixed the typos, changed the link and put the sentence about config files outside the sublist, but at the end. WDYT?

kombucha:
@not-an-aardvark I added three tests : 
- picking up a prettierrc file config
- overriding it with `[object]` option
- overriding it with `"fb"` option

I had to update the yarn.lock, otherwise tests are failing because `prettier` is locked to version 1.6.1

Hope that's ok !

eslint-plugin-prettier.js
lencioni: This makes ESLint a lot slower. I have a partial fix here, but it would be good to figure out a better way to do this: https://github.com/prettier/eslint-plugin-prettier/pull/303

PR #57, opened 2017-09-21
Allow for use of forks of prettier

hiddentao:
This change allows for forks of prettier such as [prettier-miscellaneous](https://github.com/arijs/prettier-miscellaneous) to be used instead. Unlike #15 it doesn't use any special tricks to find the specified module, and thus avoids any associated problems.

`.eslintrc` syntax demo (e.g. if you want to use `prettier-miscellaneous`):

```js
...
    "prettier/prettier": [
      "error",
      {
        bracketSpacing: true,
        jsxBracketSameLine: true,
        printWidth: 80,
        ...
      },
      null,
      "prettier-miscellaneous"
    ]
  },
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Then, in your `.eslintrc.json`:
 
 * The first option:
   - Objects are passed directly to Prettier as [options](https://github.com/prettier/prettier#options). Example:
-    
+
     ```json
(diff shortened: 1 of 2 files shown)


not-an-aardvark:
Thanks for the PR. However, as I mentioned in #15, I think this functionality is out-of-scope for this project. This is intended to be a plugin for prettier, not arbitrary libraries that expose a similar API.

PR #73, opened 2017-12-16
New: Add 'recommended' configuration

### LINKED ISSUE #72: Export a "recommended" configuration
Usually when I set up a project with Prettier I use both `eslint-{plugin,config}-prettier`.

```yaml
---
plugins:
- prettier # eslint-plugin-prettier
extends:
- prettier # eslint-config-prettier
rules:
  prettier/prettier: error # eslint-plugin-prettier
```

What would be great is if this module exported a `"recommended"` configuration that did this for me:

```yaml
---
extends:
- plugin:prettier/recommended
```

Effectively:

```js
module.exports.configs = {
  recommended: {
    extends: ['prettier'],
    plugins: ['prettier'],
    rules: {
      'prettier/prettier': 'error'
    },
  }
}
```

Would such an addition be welcome?

azz:
Closes #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ error: Delete `;` (prettier/prettier) at pkg/commons-atom/ActiveEditorRegistry.j
 ## Installation
 
 ```sh
-npm install --save-dev prettier eslint-plugin-prettier
+npm install --save-dev eslint-plugin-prettier
+npm install --save-dev --save-exact prettier
 ```
 
 **_`eslint-plugin-prettier` does not install Prettier or ESLint for you._** _You must install these yourself._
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -51,10 +51,42 @@ Then, in your `.eslintrc.json`:
...
+
+1. Enables `eslint-plugin-prettier`.
+2. Sets the `prettier/prettier` rule to `"error"`.
+3. Extends the `eslint-config-prettier` configuration.
+
+You can then set Prettier's own options inside a `.prettierrc` file.
+
 ## Options
 
+> Note: While it is possible to pass options to Prettier via your ESLint configuration file, it is not recommended because editor extensions such as `prettier-atom` and `prettier-vscode` **will** read [`.prettierrc`](https://prettier.io/docs/en/configuration.html), but **won't** read settings from ESLint, which can lead to an inconsistent experience.
```
not-an-aardvark: As a sidenote: I was under the impression that people using `eslint-plugin-prettier` would just use editor integrations for `eslint`, because it wouldn't be necessary to use integrations for `prettier`. Is it common for people to use `prettier-atom` along with `eslint-plugin-prettier`?

README.md
```diff
@@ -51,10 +51,42 @@ Then, in your `.eslintrc.json`:
...
+npm install --save-dev eslint-config-prettier
+```
+
+Then all you need in your `.eslintrc.json` is:
+
+```json
+{
+  "extends": [
+    "plugin:prettier/recommended"
+  ]
```
not-an-aardvark: Is it also necessary to use `"plugins": ["prettier"]` in this config file?

README.md
```diff
@@ -111,7 +143,9 @@ Then, in your `.eslintrc.json`:
 
 ---
 
-This plugin works best if you disable all other ESLint rules relating to code formatting, and only enable rules that detect patterns in the AST. (If another active ESLint rule disagrees with `prettier` about how code should be formatted, it will be impossible to avoid lint errors.) You can use [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to disable all formatting-related ESLint rules. If your desired formatting does not match the `prettier` output, you should use a different tool such as [prettier-eslint](https://github.com/prettier/prettier-eslint) instead.
+## Note
+
+If your desired formatting does not match the `prettier` output, you should use a different tool such as [prettier-eslint](https://github.com/prettier/prettier-eslint) instead.
```
not-an-aardvark: Nitpick: I think it would be better to keep this note next to the comments about `eslint-config-prettier`. Can you move it to the `Recommended Configuration` section (maybe under the first paragraph)?

README.md
azz: No, configurations can turn on plugins, including their own.

README.md
azz: I personally use `prettier-vscode` with `eslint-plugin-prettier`. I don't use use the `vscode-eslint` plugin's format on save capabilities as Prettier covers a wider range of languages, and I find it a bit faster.

README.md
azz: Done.

README.md
```diff
@@ -33,7 +33,8 @@ error: Delete `;` (prettier/prettier) at pkg/commons-atom/ActiveEditorRegistry.j
 ## Installation
 
 ```sh
-npm install --save-dev prettier eslint-plugin-prettier
+npm install --save-dev eslint-plugin-prettier
+npm install --save-dev --save-exact prettier
```
azz: Exact installation is recommended for `prettier` https://prettier.io/docs/en/install.html

README.md
not-an-aardvark: Hmm, I had been under the impression that `"plugins": ["prettier"]` was needed to use `plugin:prettier/recommended` at all, but after trying it out it seems like you're correct.

Separately, maybe it would be a good idea to update ESLint's plugin-loading so that `extends: plugin:foo/bar` also implies `plugins: ["foo"]`, so that it wouldn't be necessary to do something like `plugins: ["prettier"]` in a shareable config within `eslint-plugin-prettier` itself.

not-an-aardvark:
Thanks!

README.md
azz: Yeah I was surprised that it works the way it does. Not sure if there's a use case for a plugin exporting a configuration without exposing its rules.

PR #75, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can
know where their documentation is without having to resort to external packages to guess.


not-an-aardvark:
Thanks!

PR #78, opened 2018-01-16
New: Support usage alongside eslint-plugin-vue

yyx990803:
Now that Prettier core 1.10 supports Vue files, we are trying to integrate this plugin to be used alongside [eslint-plugin-vue](https://github.com/vuejs/eslint-plugin-vue), which uses `vue-eslint-parser` and allows ESLint to directly lint `*.vue` files.

To get it working together though, we need to explicitly set prettier's parser to `vue` when a vue file is encountered in this plugin. Otherwise prettier parses the vue file as JavaScript and throws syntax errors.

/cc @vjeux

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -360,19 +360,24 @@ module.exports = {
               prettier = require('prettier');
             }
 
+            const filename = context.getFilename();
             const eslintPrettierOptions =
               context.options[0] === 'fb'
                 ? FB_PRETTIER_OPTIONS
                 : context.options[0];
             const prettierRcOptions =
               prettier.resolveConfig && prettier.resolveConfig.sync
-                ? prettier.resolveConfig.sync(context.getFilename())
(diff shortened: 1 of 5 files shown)


azz:
Please see https://github.com/prettier/eslint-plugin-prettier/pull/76

not-an-aardvark:
Is this obsoleted by https://github.com/prettier/eslint-plugin-prettier/pull/76?

azz:
#76 is a more generic fix, delegating the parser decision to `prettier`.

not-an-aardvark:
Ok, I'll close this and do a release then.

yyx990803:
Ah great, that's indeed even better. Could use the tests from this one though.

not-an-aardvark:
Good idea, I fixed the merge conflict and added the test in https://github.com/prettier/eslint-plugin-prettier/commit/1ab43fd601a67100cb03bbfe614203fd399d40bb.

azz:
Thanks!

PR #83, opened 2018-01-29
Add option to skip loading prettierrc

eliperelman:
Hello! This PR adds an option to allow you to skip loading the prettierrc file. I have a use case for this in some external tooling and it would really come in handy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,11 +152,11 @@ You can then set Prettier's own options inside a `.prettierrc` file.
       }]
       ```
       
-    - `usePrettierRc`: Enables loading of the Prettier configuration file, set to `false` to disable. May be useful if you are using multiple tools that conflict with each other, or do not wish to mix your ESLint settings with your Prettier configuration.
+    - `usePrettierrc`: Enables loading of the Prettier configuration file, (default: `true`). May be useful if you are using multiple tools that conflict with each other, or do not wish to mix your ESLint settings with your Prettier configuration.
     
       ```json
       "prettier/prettier": ["error", null, {
-        "usePrettierRc": false
+        "usePrettierrc": false
       }]
       ```
 
--- a/README.md
+++ b/README.md
@@ -141,6 +141,24 @@ You can then set Prettier's own options inside a `.prettierrc` file.
     ```
 
     _This option is useful if you're migrating a large codebase and already use pragmas like `@flow`._
+  
+  - An object with the following options
+  
+    - `pragma`: Also sets the aforementioned `pragma`: a string with a pragma that triggers this rule. By default, this rule applies to all files. However, if you set a pragma (this option), only files with that pragma in the heading docblock will be checked. All pragmas must start with `@`.
+    
+      ```json
+      "prettier/prettier": ["error", null, {
+        "pragma": "@prettier"
+      }]
+      ```
+      
+    - `usePrettierRc`: Enables loading of the Prettier configuration file, set to `false` to disable. May be useful if you are using multiple tools that conflict with each other, or do not wish to mix your ESLint settings with your Prettier configuration.
+    
+      ```json
+      "prettier/prettier": ["error", null, {
+        "usePrettierRc": false
+      }]
+      ```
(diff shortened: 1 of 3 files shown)


not-an-aardvark:
Thanks for the PR. I'm fine with adding something like this, but I think it could be confusing to include it in the same object as the other prettier options, since this isn't an option that gets passed to prettier. Can you make this a second argument instead? For example, it could get configured as `prettier/prettier: [error, {}, { usePrettierrc: false }]`

eliperelman:
@not-an-aardvark absolutely, I like that route waaay better! Will revise.

eliperelman:
@not-an-aardvark updated!

eliperelman:
For posterity, this option was inspired by ESLint CLIEngine's `useEslintRc` option.

https://eslint.org/docs/developer-guide/nodejs-api#cliengine

README.md
```diff
@@ -141,6 +141,24 @@ You can then set Prettier's own options inside a `.prettierrc` file.
...
+  
+    - `pragma`: Also sets the aforementioned `pragma`: a string with a pragma that triggers this rule. By default, this rule applies to all files. However, if you set a pragma (this option), only files with that pragma in the heading docblock will be checked. All pragmas must start with `@`.
+    
+      ```json
+      "prettier/prettier": ["error", null, {
+        "pragma": "@prettier"
+      }]
+      ```
+      
+    - `usePrettierRc`: Enables loading of the Prettier configuration file, set to `false` to disable. May be useful if you are using multiple tools that conflict with each other, or do not wish to mix your ESLint settings with your Prettier configuration.
```
not-an-aardvark: Minor nitpick: Can we rename the option to `usePrettierrc` (lowercase `r`)? That would also be consistent with the `useEslintrc` option that this is based on.

not-an-aardvark: > set to `false` to disable

Nitpick: Can you replace this text with "(default: `true`)"? The double negative in "set to `false` to disable" makes it a bit confusing to read.

eslint-plugin-prettier.js
```diff
@@ -313,15 +331,26 @@ module.exports = {
...
-          { type: 'string', pattern: '^@\\w+$' }
+          {
+            anyOf: [
+              // Pragma:
+              { type: 'string', pattern: '^@\\w+$' },
+              {
+                type: 'object',
+                properties: {
+                  pragma: { type: 'string', pattern: '^@\\w+$' },
+                  usePrettierRc: { type: 'boolean' }
```
not-an-aardvark: Can you add a test for the new option? (It would probably be similar to the existing `prettierrc` tests, except that the test would be asserting that the options in the file do not get applied.)

eliperelman:
@not-an-aardvark ok, I think all your requests have been made! I also added a test to ensure you could still use `pragma` from the options object as well.

azz:
If this is a breaking change (moving `pragma` from first object to second), I'd recommend removing it entirely as it is now built-in to Prettier. 

Perhaps a more ergonomic API would be to do:

```js
"prettier/prettier": ["error"], // use config file
"prettier/prettier": ["error", {}], // do not use config file
"prettier/prettier": ["error", { "singleQuote": true }], // do not use config file
```

not-an-aardvark:
I think this is backwards-compatible -- `pragma` was already in the second slot. (I agree that it should probably be removed in the next major release.)

eliperelman:
This is not a breaking change, it's a minor bump. Using a string pragma is still supported in the second slot.

azz:
Yeah I misinterpreted the change, ignore the first part of my comment.

not-an-aardvark:
If I'm understanding correctly, the second part of the comment would be a breaking change, right? Maybe we should merge this as-is and then implement the second part of the comment in a major release (at some indefinite point in the future if a major release is necessary).

not-an-aardvark:
Thanks for contributing!

eliperelman:
My pleasure, thank you for accepting!

PR #96, opened 2018-06-25
Add clarification notice about Flow/React support to the readme

### LINKED ISSUE #95: Add clarification notice about Flow/React support to the readme
In the readme’s “Recommended Configuration” section, there are no instructions about the need to add e.g. `”extends”: [”prettier/react”]` to support its corresponding ESLint plugin.

I think there should be at least a clarification notice about this for disabling rules of special ESLint plugins (e.g. `eslint-plugin-react`).

Example: https://github.com/kripod/gatsby-starter-modern/blob/cabc047185ffdfe27c29d31071999d80fe39b474/.eslintrc.json#L3

kripod:
Resolves #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ To integrate this plugin with `eslint-config-prettier`, you can use the `"recomm
   npm install --save-dev eslint-config-prettier
   ```
 
-2. Then all you need in your `.eslintrc.json` is:
+2. Then you need to add `plugin:prettier/recommended` as the last extension in your `.eslintrc.json`:
 
   ```json
   {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +84,21 @@ This does three things:
 
 You can then set Prettier's own options inside a `.prettierrc` file.
 
+3. In order to support special ESLint plugins (e.g. [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react)), add extra exclusions for the plugins you use like so:
```
not-an-aardvark: I think the numbering here is a bit confusing because there is a separate numbered list directly above this line. Can you change the list above this one to use bullet points (by prefixing each line with `*` rather than numbers?

kripod: Sure, I’ve changed it.

not-an-aardvark:
Thanks!

PR #100, opened 2018-07-05
Add representation for \r to showInvisibles

### LINKED ISSUE #99: Add carriage return "\r" to showInvisibles
Prettier will format files using the first line-ending it sees. This means that files that use 
\r\n as line breaks will keep them.

showInvisibles has conversions for space, \n and \t. It'd be good if it had a replacement for \r as well so report output doesn't look goofy for crlf files as the terminal tries to render the literal carriage return.

Unfortunately I'm not sure what the best solution for this is. There's no character that screams "I'm a carriage return" that I'm aware of. Perhaps ⏎ can be used for \n (as it is now), and a slightly different character can be used for a \r\n combo?

I reckon we can ignore files with just \r line endings (macos 9 lols)

BPScott:
Fixes #99


BPScott:
Build failure seems to be unrelated to the changes in this PR

Issue #88, opened 2018-04-09
prettierignore doesnt function at all


Mr-Chilly:
"eslint": "^3.10.0",
"prettier": "^1.11.1"
"eslint-plugin-prettier": "^2.6.0",

**What did you expect to happen?**
prettierignore to blacklist files in the project

**What actually happened?**
.prettierignore has no effect, only .eslintignore. This means that code I want to lint, but not prettify isn't possible

not-an-aardvark:
Hi, thanks for creating an issue.

Would it work to conditionally disable the rule from `eslint-plugin-prettier` by using [ESLint's glob configuration](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns)? This seems like it would avoid the need to have a `.prettierignore` file. (Note that glob configuration is only supported starting from ESLint 4.1.0.

Mr-Chilly:
Thanks for the reply! That definitely seems like an option, however it would have the problem of not being shareable with other prettier plugins. I guess it has all the same benefits of the plugin using .prettierrc config files

davidlygagnon:
I'm seeing the same issue. The repos I'm working with contains both an .eslintignore and .prettierignore. I'm trying to consolidate prettier + eslint with eslint-plugin-prettier but the files ignores in .prettierignore are not ignored. If eslint-plugin-prettier allowed specifying an ignore file, that would make things easier.

heisian:
`.prettierignore` works with the prettier CLI, so I don't understand why it wouldn't be supported for this plugin.

wlingke:
@not-an-aardvark Can you provide an example for how to do what you suggested? Turning `prettier/prettier: "off"` doesn't stop prettier from formatting the file. Only `prettier-ignore` does.

not-an-aardvark:
Disabling the `prettier/prettier` rule will prevent `eslint-plugin-prettier` from formatting your files, but it won't prevent `prettier` from formatting the files if you run `prettier` out-of-band separately from this plugin.

heisian:
got it, thanks @wlingke and thanks @not-an-aardvark for clearing it up.

Issue #98, opened 2018-07-02
Override parser to babylon - Fix compatibility with eslint-plugin-html and prettier@1.13.x


nappy:
Upgrading prettier from ~1.12.1 to ^1.13.6 causes eslint-plugin-html to break.

The use case is, to lint and format javascript inside html script tags.

I originally posted the issue at https://github.com/prettier/prettier/issues/4759, but it became clear that it should be fixed here.

Prettier 1.13 changes how the parser is inferred, previously it would default to `babylon`, but now this must be explicitly defined in the `.prettierrc` config.
Since `eslint` only lints javascript it would make sense to restore the previous behavior by overriding the parser to `babylon` when calling the prettier API (if possible), and therefore restore compatibility with previous working configurations.

Please read the referenced issue, if its not clear from this description.

**Environments:**
- Prettier Version: 1.13.6
- Running Prettier via: eslint-plugin-prettier
- Runtime: Node v10.5.0 Yarn 1.7.0 Npm 6.1.0
- Operating System: Windows
eslint@^5.0.1
eslint-plugin-html@^4.0.5
eslint-plugin-prettier@^2.6.1
prettier@^1.13.6

**Steps to reproduce:**
Add plugins html and prettier to .eslintrc
Add an html file to your project
Run eslint **/*.{js,html}

**Expected behavior: (with prettier 1.12.1)**
yarn run v1.7.0
$ eslint **/*.{js,html}
Done in 5.89s.

**Actual behavior: (with prettier 1.13.6)**
yarn run v1.7.0
$ eslint **/*.{js,html}
No parser could be inferred for file: C:\Code\eslint-html\assets\index.html
Error: No parser could be inferred for file: C:\Code\eslint-html\assets\index.html
    at normalize (C:\Code\eslint-html\node_modules\prettier\index.js:7058:15)
    at formatWithCursor (C:\Code\eslint-html\node_modules\prettier\index.js:10378:12)
    at C:\Code\eslint-html\node_modules\prettier\index.js:31217:15
    at Object.format (C:\Code\eslint-html\node_modules\prettier\index.js:31236:12)
    at Program (C:\Code\eslint-html\node_modules\eslint-plugin-prettier\eslint-plugin-prettier.js:412:45)
    at listeners.(anonymous function).forEach.listener (C:\Code\eslint-html\node_modules\eslint\lib\util\safe-emitter.js:45:58)
    at Array.forEach (<anonymous>)
[...]

PR #104, opened 2018-07-17
Respect .prettierignore config

kapowaz:
This PR fixes an issue with eslint-plugin-prettier not respecting file ignore patterns specified in `.prettierignore` (#88).


eslint-plugin-prettier.js
```diff
@@ -401,48 +402,52 @@ module.exports = {
...
-                      difference.deleteText,
-                      difference.insertText
-                    );
-                    break;
-                }
-              });
-            }
+
+            prettier.getFileInfo(sourceFileName, {
+              ignorePath: '.prettierignore'
```
kapowaz: I’m not sure if there’s some generic mechanism for exposing default config so that people can specify `eslint-plugin-prettier` to use a different file than `.prettierignore`, but given that this was broken until now I’m not sure just how much it matters… maybe one to be addressed in a later PR? 🤷‍♂️

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.6.3 (2018-07-17)
+
+* Fix: respect .prettierignore settings ([#88](https://github.com/prettier/eslint-plugin-prettier/issues/88))
+
```
not-an-aardvark: It's not necessary to update the changelog -- it will be automatically updated during the release.

eslint-plugin-prettier.js
```diff
@@ -401,48 +402,55 @@ module.exports = {
...
+            }).then((sourceFileInfo) => {
```
not-an-aardvark: I don't think this will work -- ESLint rules are required to be synchronous. Does prettier provide a synchronous API for this?

CHANGELOG.md
kapowaz: Great! I did wonder if that might be automated… 🙈

eslint-plugin-prettier.js
kapowaz: Yes! I tried both ways so I can update this to use the `.sync` version 👍

kapowaz:
Thanks for the comments — I’ve pushed a revised PR, hopefully this should address the issues. The tests still fail due to the issues mentioned in #98, but assuming that can be fixed I think this should be good.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-prettier",
-  "version": "2.6.2",
+  "version": "2.6.3",
```
not-an-aardvark: Sorry, I should have mentioned this before, but it's also not necessary to update the version in `package.json` (it will also be automatically updated on release).

eslint-plugin-prettier.js
```diff
@@ -401,19 +402,22 @@ module.exports = {
...
               {},
               prettierRcOptions,
               eslintPrettierOptions,
-              { filepath: context.getFilename() }
+              { filepath: sourceFileName }
             );
-
             const prettierSource = prettier.format(source, prettierOptions);
-            if (source !== prettierSource) {
+            const sourceFileInfo = prettier.getFileInfo.sync(sourceFileName, {
```
not-an-aardvark: Can you add a test for this behavior?

not-an-aardvark: Can you check for the existence of `prettier.getFileInfo.sync` before calling it? The plugin needs to continue being able to work for users on old versions of `prettier`.

kapowaz: I’ll look into adding a test for this.

As regards checking for the existence of `prettier.getFileInfo.sync` I could do this, but this functionality is required in order to test for whether a file is being ignored,  and so I’ve bumped the dependency to prettier `^1.13.0` for that reason. Is there a reason even future releases of `eslint-plugin-prettier` _needs_ to always support old versions of prettier?

Maybe this be a minor release instead of a patch release, perhaps?

not-an-aardvark: Since the user installs `prettier` independently of `eslint-plugin-prettier`, it would be a breaking change to drop support for old versions of `prettier`, because we don't have any control over what version the user has installed.

kapowaz: I assume that’s an intentional thing, rather than explicitly having this depend on a given version of prettier? I’m not sure it would be possible to add a test to cover checking for a file being correctly ignored if we can’t guarantee a minimum version of prettier is installed though. Would welcome ideas if you think that is possible!

kapowaz:
I’ve updated the PR again with some existence-guarding checks against `prettier.getFileInfo`. As I commented above, I’m not sure how we can reliably test for this functionality if there is no specific version of prettier guaranteed to be installed (and `getFileInfo` was only added in v1.13.0). Feedback on how to solve that welcomed!

eslint-plugin-prettier.js
not-an-aardvark: Yes, the reasoning is that the user might want to choose a specific version of `prettier` to lint their code with (also see: https://github.com/prettier/eslint-plugin-prettier/pull/1). Admittedly, this was probably more important at the time that pull request was created than it is now, because prettier wasn't yet stable and was making breaking changes more frequently, and people wanted to use the plugin with several different versions.

kapowaz: Cool. Like I say this can’t be merged until #98 is fixed anyway, so I guess I can revisit when that is handled (or maybe dive in and have a go at fixing that myself! 😁)

not-an-aardvark: Sure, that would be great.

PR #111, opened 2018-09-26
Support prettierignore and custom processors

### LINKED ISSUE #81: Error when used with eslint-plugin-graphql in v2.5.0
**What version of `eslint` are you using?**
4.15.0

**What version of `prettier` are you using?**
1.10.2

**What version of `eslint-plugin-prettier` are you using?**
2.5.0

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
Relevant rules: 
```
{
    'prettier/prettier': [
      1,
      {
        printWidth: 120,
        trailingComma: 'es5',
        singleQuote: true,
        semi: false,
      },
    ],
    'graphql/template-strings': [
      2,
      {
        env: 'literal',
        schemaJson: require('./graphql/schema.json'),
      },
    ],
}

**What source code are you linting?**
This seems to happen for any `.graphql` file while  using `eslint-plugin-graphql` (I'm on version 1.4.1).

It should lint correctly.

**What actually happened?**
In `eslint-plugin-prettier` `2.4.0` everything works, but since `2.5.0` I get the following error: 
```
Syntax Error: Unexpected Name "ESLintPluginGraphQLFile" (1:1)
> 1 | ESLintPluginGraphQLFile`#import "./redacted-file-name.graphql"
    | ^
```

Assuming this has to do with the plugin now passing the filename to prettier, and eslint-plugin-graphql causes graphql files to have a weird filename reported somehow?

### LINKED ISSUE #88: prettierignore doesnt function at all

### LINKED ISSUE #98: Override parser to babylon - Fix compatibility with eslint-plugin-html and prettier@1.13.x

### PR DESCRIPTION
**Pull request by BPScott**
Fixing two bugs as one provides the scaffolding of the fix for the other.

---

Ignore files in .prettierignore 

Fixes #88

---

Force the babylon parser when parsing not-js files 

Forcing the parser stops errors that are caused by trying to run a JS
fragment through the graphql / markdown parsers.

The 'html' parser has not yet been released yet, but will be in Prettier
1.15.

This uses a block list over an allow list because I expect the list of
"file types with a prettier parser that could contain javascript fragments"
will grow at a slower pace than "file types with a prettier parser that
are variations of the javascript language".

Fixes #98, Fixes #81

### CODE REVIEW

File: test/prettier.js [hunk 1]
```diff
@@ -65,6 +65,11 @@ ruleTester.run('prettier', rule, {
       code: `var foo = {bar: 0};\n`,
       filename: getPrettierRcJsFilename('bracket-spacing'),
       options: [{ bracketSpacing: false }, { usePrettierrc: false }]
+    },
```
not-an-aardvark: Can you add a test for the parser blacklist behavior? We don't need to actually include processors, but maybe one way to do it would be to put JavaScript code in a file with the `graphql` extension, and assert that no parsing errors are produced (since the text should be parsed as JavaScript).

eslint-plugin-prettier.js
```diff
@@ -397,19 +398,58 @@ module.exports = {
...
+            // based off the filename. Otherwise, for instance, on a .md file we
+            // end up trying to run prettier over a fragment of JS using the
+            // markdown parser, which throws an error.
+            // If we can't infer the parser from from the filename, either
+            // because no filename was provided or because there is no parser
+            // found for the filename, use javascript.
+            // This is added to the options first, so that
+            // prettierRcOptions and eslintPrettierOptions can still override
+            // the parser.
+            const parserBlocklist = [null, 'graphql', 'markdown', 'html'];
```
not-an-aardvark: This list seems like it might plausibly need to expand in the future. The detailed comment above it with background is definitely appreciated, but since the situation is fairly complex and difficult to comprehend, is there a simple description we can add for what the list should contain for future reference?

For example, maybe we could add a comment stating that:

> This list should contain the list of `prettier` parser names for file types where:
>
> * Prettier supports parsing the file type
> * There is an ESLint processor that extracts JavaScript snippets from the file type.

test/prettier.js
BPScott: Good call. I tried doing this with including a plugin, but I couldn't get the plugin to be applied within a RuleTester. Forcing the extension is a good middle ground

PR #112, opened 2018-09-27
Prettier formatting and Travis updates

### LINKED ISSUE #97: Test against ESLint 5
Hi!

Currently the Travis configuration for this repo does:
```yaml
env:
  - ESLINT_VERSION=latest
  - ESLINT_VERSION=3.15.0
before_script:
  - if [[ $ESLINT_VERSION != "latest" ]]; then
      yarn upgrade "eslint@$ESLINT_VERSION";
fi
```

And in package.json:
```js
  "devDependencies": {
    "eslint": "^3.14.1",
```

At first glance it might appear that `eslint-plugin-prettier` is being tested against "latest" (ie 5.x) and `3.15.0`. However given the `package.json` eslint version is out of date, the plugin is actually only being tested against ESLint `3.19.0` and `3.15.0` - meaning there is no test coverage of 4.x or 5.x.

I would suggest either:
1. Enabling something like https://renovatebot.com/ for this repository (so the `package.json` version is actually "latest"), and renaming `"latest"` in `.travis.yml` to something more accurate, like `"default"`.
2. Changing the `before_script` `.travis.yml` section to remove the conditional, so that it always runs `yarn upgrade "eslint@$ESLINT_VERSION"`, including in the `"latest"` case.

I think option (1) seems preferable, since Renovate would also ensure other dependencies are kept up to date -- plus (2) would be confusing, since someone just looking at `package.json` (and not also `.travis.yml`) might mistakenly think there was still no test coverage of more recent ESLint versions.

BPScott:
Chore: Add format script to run prettier

Run the format script and commit updates

---

Chore: Improve travis matrix

Ensure we test with eslint 3.x, 4.x and 5.x, and node 10

Fixes #97

---

Chore: Add vscode config to autoformat on save

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,5 @@
 package.json
+CHANGELOG.md
 
 # this file doesn't exist, but we use it as a filename that should be ignored
 # by prettier in the tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,31 @@
 language: node_js
 cache: yarn
 node_js:
-  - "4"
-  - "6"
-  - "7"
-  - "8"
+  - "10"
   - "9"
+  - "8"
+  - "7"
+  - "6"
+  - "4"
 env:
-  - ESLINT_VERSION=latest
-  - ESLINT_VERSION=3.15.0
-before_script:
-  - if [[ $ESLINT_VERSION != "latest" ]]; then
+  - ESLINT_VERSION=current
+  - ESLINT_VERSION=^5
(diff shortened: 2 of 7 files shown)


BPScott:
Mashing a few things into a single PR as it's a little easier than spinning out multiple PRs.

.prettierignore
```diff
@@ -1,4 +1,5 @@
 package.json
+CHANGELOG.md
```
BPScott: Don't touch autogenerated files

Issue #54, opened 2017-09-13
Transition plan for new `requirePragma` in prettier core


wbinnssmith:
I just got the pragma functionality merged into prettier itself at prettier/prettier#2772. It is very similar to the functionality implemented here, only:
  * Now editors can use their autoformat functionality and follow the same rules for applying based on the pragma as other tools
  * The name of the pragma is fixed to either `@prettier` or `@format`. There's a bit of discussion in that PR behind this reasoning.

How should we manage the transition from `eslint-plugin-prettier`'s implementation of pragmas to the new one in core? Off the top of my head it makes sense to no longer use the third option for this plugin, and instead either rely on a `.prettierrc` or instruct users to pass along `requirePragma` in the second options argument of this plugin.

Thanks! cc @asuarez

not-an-aardvark:
I think we could probably deprecate the pragma option in `eslint-plugin-prettier` in favor of using `prettier`'s version directly. People won't be able to use pragmas other than `@prettier` or `@format`, but my impression is that being able to customize the pragma isn't a very important feature.

BPScott:
Closed as pragma functionality has been removed in favour of using prettier's native pragma support as of v3

Issue #79, opened 2018-01-17
Bad formatting closing script-tag in vue-files


Alex-Sokolov:
**What version of `eslint` are you using?**
`"eslint": "4.15.0",`

**What version of `prettier` are you using?**
`"prettier": "1.10.2",`

**What version of `eslint-plugin-prettier` are you using?**
`"eslint-plugin-prettier": "2.5.0",`

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
```
const resolve = require('path').resolve;

module.exports = {
  root: true,
  parserOptions: {
    parser: 'babel-eslint'
  },
  env: {
    browser: true,
    node: true,
    jest: true
  },
  extends: [
    'standard',
    'prettier',
    'plugin:vue/base',
    'plugin:vue/essential',
    'plugin:vue/strongly-recommended',
    'plugin:vue/recommended'
  ],
  plugins: ['html', 'variables', 'prettier'],
  rules: {
    'prettier/prettier': 'error',
    'import/no-unresolved': 'error',
    'variables/only-ascii-variables': 'error'
  },
  settings: {
    'import/resolver': {
      webpack: {
        config: {
          resolve: {
            alias: {
              '~': __dirname
            }
          }
        }
      }
    }
  }
};
```

```
"prettier": {
    "printWidth": 120,
    "tabWidth": 2,
    "useTabs": false,
    "semi": true,
    "singleQuote": true,
    "trailingComma": "none",
    "bracketSpacing": true
  },
```

**What source code are you linting?**
https://pastebin.com/Hzik4pNs

**What did you expect to happen?**
Formatting correct

**What actually happened?**
After updating `eslint-plugin-prettier` from 2.4.0 to 2.5.0 and run `eslint --fix --ext .js,.vue --ignore-path .gitignore .`
![image](https://user-images.githubusercontent.com/4497128/35028877-3ddb3bee-fb69-11e7-91d3-372638256ee1.png)

azz:
[...]

PR #113, opened 2018-09-27
V3

### LINKED ISSUE #48: Relax pragma rules
At the moment the pragma has to be the first line and in the `/** @yourPragma */` format. If it's used in a file with Flow it means you have to remember to put the Prettier one first.

This isn't a huge hassle but it caused me a few minutes of confusion as I thought I had things set up wrong. It would be nice if this plugin's pragma were more like Flow's and could be in either `// @yourPragma` or `/** @yourPragma */` format and be anywhere in the file.

BPScott:
Version 3 takes an opportunity to clear up some logic, by updating the minimum required node version (dropping support for v4, v7 and v9), eslint version (dropping support for v3 and v4) and prettier version (requiring at least v1.13.0).

Most notably this allows us to remove pragma support and tell people to rely upon prettier's native pragma support (with a side effect of no longer needing to depend upon jest-docblock), which fixes #48 and #54.

We also drop support for the "fb" option which allows us to simplify the options passed into the rule. Getting people to use a .prettierrc is now the preferred approach.

The `showInvisibles` and `generateDifferences` have been moved into a [new package](https://github.com/prettier/prettier-linter-helpers) so stylelint-prettier and tslint-plugin-prettier don't have to have a runtime dependency on eslint-plugin-prettier as they consume those helper functions.

This updates a few dev dependencies along the way (which could have been done as v2.x but I was on a roll)


.travis.yml
```diff
@@ -4,9 +4,7 @@ node_js:
   - "10"
   - "9"
   - "8"
-  - "7"
   - "6"
-  - "4"
 env:
```
not-an-aardvark: If we're dropping support for Node 4 and 7, should we also drop support for some old versions of ESLint? There's currently [some logic](https://github.com/prettier/eslint-plugin-prettier/blob/9fac6b4039c1983b83073fa7af7864f0d7e1f2d3/eslint-plugin-prettier.js#L51-L83) that only runs on versions of ESLint before `3.16.0`.

.travis.yml
```diff
@@ -4,9 +4,7 @@ node_js:
   - "10"
   - "9"
```
not-an-aardvark: Should we also drop Node 9 now that it's EOL?

.travis.yml
not-an-aardvark: I have no strong preferences either way. I suppose we might as well drop support for 4.x (there aren't many breaking changes in ESLint v5, so hopefully it shouldn't be too difficult for users to upgrade).

.travis.yml
BPScott: I was torn as ESLint v5 left node v9 in its test matrix, but I think that was because ESLint v5 was released JUST after node v9's EOL. (ESLint v5 release 2018-06-22, Node v9 was EOLed 2018-06-30).

I'm happy to drop support for it if you are.

not-an-aardvark: Sounds good to me, I don't think we need to optimize for users running EOL versions of Node.

zertosh:
@vjeux, I think it's time to finally have a global `.prettierrc` - wdyt?

vjeux:
Prettier now uses cosmiconfig and respects .prettierrc

BPScott:
@zertosh: Can you define "global"? As @vjeux says, Prettier already supports a `.prettierrc` that can sit within a project at the same level as `node_modules`. and contain:

```json
{
   "singleQuote": true,
   "trailingComma": "all",
   "bracketSpacing": false,
   "jsxBracketSameLine": true,
   "overrides": [
    {"files": "*.js", "options": {"parser": "flow"}}
  ]
}
```

And those options will be honoured by `node_modules/.bin/prettier some-file.js` or `node_modules/.bin/prettier eslint some-file.js`.

Thanks to cosmiconfig you can also name that file `.prettierrc.js` and have it `require()` a shared configuration you provide from elsewhere.

zertosh:
@BPScott, sorry about that comment. I should've DM @vjeux directly (which I did after his comment). That was in reference to the way we run Prettier at Facebook. This change is fine, we'll just have to make some adjustments on our end.

BPScott:
Gotcha :)

So to confirm: I'm good to go ahead and drop support for [the "fb" option](https://github.com/prettier/eslint-plugin-prettier/blob/master/README.md#options) and make it so you can only pass in an object representing a prettier object (or null)?

zertosh:
@BPScott, yup!

azz:
This simplifies this module significantly, nice!

azz:
One thing worth considering: removing the optional prettier config in eslintrc, and only supporting prettierrc. The reason for this is editor integrations like prettier-vscode, prettier-atom do not read eslint config files

azz:
Gah, typing on mobile, sorry.

BPScott:
We should certainly update the docs to push for "we read prettierrc, you shouldn't need to add custom config in the majority of cases".

However I think it is worth keeping the config override because eslint's processor integrations means that it can potentially read not-js, parse the js files out of them and then pass those js fragments to prettier, and we need a way to force only-prettier-within-eslint to parse those file using not the standard parser for that file type. We can cover the common use-cases but there's no way to have correct behaviour for everybody.

You can see that headache in action in #79

BPScott:
Yo @not-an-aardvark, I've added a few more things since your initial review. Could you check them over and confirm you're happy with them? They are: dropping support for ESLint 3/4 and extracting two of the utility functions into a separate package so stylelint-prettier and tslint-plugin-prettier don't need to depend upon eslint-plugin-prettier. Check the commit messages for more info.

I think this is it for code changes in v3. If you're happy then this this can be merged and released as v3.0.0.

not-an-aardvark:
Published as v3.0.0!

PR #118, opened 2018-10-10
New: option `withNodeModules` to unignore `node_modules` folders.

reklatsmasters:
This PR add option `withNodeModules` (default: `false`) to unignore `node_modules` folders. May be useful if you are using multiple `node_modules` directories, example to avoid many `../..` sequences.

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -142,7 +142,7 @@ module.exports = {
         const usePrettierrc =
           !context.options[1] || context.options[1].usePrettierrc !== false;
         const withNodeModules =
-          !context.options[1] || context.options[1].withNodeModules !== false;
+          context.options[1] && context.options[1].withNodeModules !== false;
         const sourceCode = context.getSourceCode();
         const filepath = context.getFilename();
         const source = sourceCode.text;
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -131,7 +131,8 @@ module.exports = {
           {
             type: 'object',
             properties: {
-              usePrettierrc: { type: 'boolean' }
+              usePrettierrc: { type: 'boolean' },
+              withNodeModules: { type: 'boolean' }
             },
             additionalProperties: true
           }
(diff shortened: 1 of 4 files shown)


BPScott:
Hi @reklatsmasters, can you please add a test for this

eslint-plugin-prettier.js
```diff
@@ -164,7 +167,8 @@ module.exports = {
               : null;
 
             const prettierFileInfo = prettier.getFileInfo.sync(filepath, {
-              ignorePath: '.prettierignore'
+              ignorePath: '.prettierignore',
+              withNodeModules
```
BPScott: Rather than pass in a single value within the options here, I think it would be more future-compatible to pass in an object that is merged with some defaults. That means that if prettier decides to add additional options to getFileInfo() then we won't need to make additional changes to accommodate them.

Something like this (written but not tested):

```js
const fileInfoOptions = Object.assign(
    {},
    { ignorePath: '.prettierignore' }
    eslintFileInfoOptions
)

const prettierFileInfo = prettier.getFileInfo.sync(filepath, fileInfoOptions);
```

Where `eslintFileInfoOptions` is defined above as something like this: (written but not tested)

```js
const eslintFileInfoOptions = context.options[1] && context.options[1].fileInfoOptions || {}
```

eslint-plugin-prettier.js
```diff
@@ -131,7 +131,8 @@ module.exports = {
           {
             type: 'object',
             properties: {
-              usePrettierrc: { type: 'boolean' }
+              usePrettierrc: { type: 'boolean' },
+              withNodeModules: { type: 'boolean' }
```
BPScott: Make this an object named `fileInfoOptions`

test/prettier.js
```diff
@@ -61,6 +61,17 @@ ruleTester.run('prettier', rule, {
...
+    // Should ignore files from node_modules
+    {
+      code: 'a();;;;;;\n',
+      filename: 'node_modules/dummy.js'
+    },
+    // Should check files from node_modules too
+    {
+      code: 'a();\n',
+      filename: 'node_modules/dummy.js',
+      options: [{}, { withNodeModules: true }]
```
BPScott: Updated usage would be 

```
options: [{}, { fileInfoOptions: { withNodeModules: true }}]
```

BPScott:
Replaced by #187

PR #128, opened 2018-11-14
Move eslint-config-prettier to peerDependencies

lucaslugao:
The  eslint-config-prettier package should be a peer dependency instead of a dev dependency. This guarantees that eslint-config-prettier is available to be extended by the parent's eslint configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,13 +32,13 @@
   },
   "peerDependencies": {
     "eslint": ">= 5.0.0",
+    "eslint-config-prettier": "^3.3.0",
     "prettier": ">= 1.13.0"
   },
   "devDependencies": {
     "@not-an-aardvark/node-release-script": "^0.1.0",
     "eslint": "^5.6.0",
     "eslint-config-not-an-aardvark": "^2.1.0",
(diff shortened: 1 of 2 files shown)


j-f1:
IMO it should stay as it is now. Since `devDependencies` aren’t installed by default, there isn’t a need to remove it from there. Also, since the config isn’t a `peerDepenency`, it doesn’t need to be installed if you’re not going to use it.

If you feel differently about this, I’m happy to discuss!

lucaslugao:
Thanks for the answers! In the PR, I mistakenly deleted the `eslint-config-prettier` from the devDependencies. Later, I realized that I only wanted to add it as peerDependency too (as you probably understood).

Thanks for making it clear that the choice not to warn the user about this was intentional. I was only concerned about having for example to dig into the eslint output in vscode to find out that the `eslint-config-prettier` was not being found. 

After rereading the readme of this package I realised that it is sufficiently clear that the rule `plugin:prettier/recommended` should be used "to integrate this plugin with `eslint-config-prettier`". It was my fault that I haven't noticed it was already being mentioned in the readme.

Thank you guys!

PR #142, opened 2018-12-28
Fixes pluginSearchDirs w/ getFileInfo

arcanis:
This plugin wasn't properly forwarding the `pluginSearchDirs` option to Prettier, causing unexpected issues in some cases. This diff simply makes sure that the options are merged with the rc settings that are read right before.

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -163,9 +163,15 @@ module.exports = {
                 })
               : null;
 
-            const prettierFileInfo = prettier.getFileInfo.sync(filepath, {
-              ignorePath: '.prettierignore'
-            });
+            const prettierFileInfo = prettier.getFileInfo.sync(
+              filepath,
+              Object.assign(
+                {
+                  ignorePath: '.prettierignore'
+                },
+                prettierRcOptions
+              )
+            );
 
             // Skip if file is ignored using a .prettierignore file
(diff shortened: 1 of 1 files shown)


BPScott:
Thanks @arcanis,

I think this should also include any options added within the eslint config - the `eslintPrettierOptions` object, can update your Object.assign to be:

```js
Object.assign(
  { ignorePath: '.prettierignore' },
  prettierRcOptions,
  eslintPrettierOptions
)
```

Would it be possible to add a test for this as well?

BPScott:
Taking a deeper look into this, I don't think this code is correct.

The options that [`prettier.getFileInfo`](https://prettier.io/docs/en/api.html#prettiergetfileinfofilepath-options) accepts are not the options that `prettier.format` accepts, so it doesn't make sense to merge the options that we pass to `format` into `getFileInfo`.

I've just merged #187 allows you to pass arbitrary options to `getFileInfo`. If you need to pass extra options into getFileInfo then you should use that.

BPScott:
It looks like getFileInfo doesn't ever use `pluginSearchDirs` (per https://github.com/prettier/prettier/blob/3fcf69a7d561a22314c9644b4e309907310db266/src/common/get-file-info.js) so if you need that you'll have to raise an issue in prettier itself.

Issue #197, opened 2019-05-27
Allow to pass a custom version of prettier as a dependency


onigoetz:
**What version of `eslint` are you using?**
latest, 5.16.0

**What version of `prettier` are you using?**
latest, 1.17.1

**What version of `eslint-plugin-prettier` are you using?**
latest, 3.1.0

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**

I use a configuration tool, Crafty, that handles that for me, but related to prettier my config is : 

```
rules: {
    "prettier/prettier": ["warn", { parser: "babel" }],
}
```

**What source code are you linting?**
N/A

**What did you expect to happen?**
The code to be linted

**What actually happened?**

```
./src/index.ts
Error: Cannot find module 'prettier'
Occurred while linting .../src/index.ts:1
    at Array.forEach (<anonymous>)
    at Array.forEach (<anonymous>)
```

So, yes, the problem is obvious, I don't have Prettier installed.
The thing is, I'm the maintainer of a tool that bundles all the tools you need to build/lint code in a single dependency (similar to create-react-app) .

Now I would like to be able to pass the full path of the prettier module as a configuration option, this way I can add prettier as a dependency. Same as you would do for a Babel Preset or other tool's configuration.
In the case of a simple user configuration it's not necessary but in case of Yarn PNP or simply a configuration that is inherited, it's better to have everything bundled.

I can provide a pull-request if necessary

not-an-aardvark:
I'm confused about why a `peerDependency` causes a problem here. It seems like if you install both `eslint-plugin-prettier` and `prettier`, then `eslint-plugin-prettier` will be able to load `prettier` since it's specified as a `peerDependency`. Is that not the case?

onigoetz:
in the case of Yarn PNP it works only if the dependency is at the top level.
For example

Standard usage : 

```
- my app
  - prettier
  - eslint-plugin-prettier
```

This will work fine as Yarn PNP will see the top level dependency and resolve it for eslint-plugin-prettier

Plugin bundled:

```
- my app
  - prettier
  - my-eslint-plugin-bundle
    - eslint-plugin-prettier
```
In this case it will work as well as prettier is still at the top level.

Plugin and Prettier bundled:

```
- my app
  - my-eslint-plugin-bundle
    - eslint-plugin-prettier
    - prettier
```

[...]

PR #199, opened 2019-05-31
Allow to specify the path to Prettier

onigoetz:
Hello @not-an-aardvark and @BPScott ,

as explained in #197, in some cases, having Prettier as a peer dependency is not ideal, for example in a mulitmodule project or with Yarn PNP.

This pull request allows, for those who need it, to specify the full path to the prettier package with `require.resolve("prettier")` from the package that has it as a dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,8 @@ For the list of every available exclusion rule set, please see the [readme of es
       }]
       ```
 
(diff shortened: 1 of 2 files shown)


BPScott:
Closing this as #197 was also closed

PR #205, opened 2019-06-18
Update: Rename babylon parser to babel

robertoentringer:
Prevent message : { parser: "babylon" } is deprecated; we now treat it as { parser: "babel" }.

Related : https://github.com/prettier/prettier/pull/5647

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -210,7 +210,7 @@ module.exports = {
             if (
               parserBlocklist.indexOf(prettierFileInfo.inferredParser) !== -1
             ) {
-              initialOptions.parser = 'babylon';
+              initialOptions.parser = 'babel';
             }
 
             const prettierOptions = Object.assign(


BPScott:
Our minimum supported version of prettier is [`1.13.0`](https://github.com/prettier/eslint-plugin-prettier/blob/60bb22f4ae1d6af001ba55338a7cb08111db23d9/package.json#L35), while the babel parser name was introduced in 1.16.0.

In order to not break people using 1.13.0 to 1.15.last, This will require either:

- Bumping our minimum supported version of prettier, which is technically a breaking change and thus requiring a major version bump, so uugh not a fan of that.
- Run [`prettier.getSupportInfo()`](https://prettier.io/docs/en/api.html#prettiergetsupportinfo-version) to set the parser to babel or babylon according to what is supported by the current version.

Can you take a stab at the later.

robertoentringer:
@BPScott thank you very much for your feedback!

robertoentringer:
Just for information purposes, by setting the parser to babel in **.eslintrc.js** rules the warning disappears.

>rules: { "prettier/prettier": ... parser: "babel" ...} 

```js
module.exports = {
  env: { browser: true, es6: true },
  extends: ["eslint:recommended", "plugin:prettier/recommended"],
  parserOptions: { ecmaVersion: 2018 },
  plugins: ["prettier", "html"],
  rules: {
    "prettier/prettier": ["error", { semi: false, printWidth: 100, parser: "babel" }],
    "comma-dangle": ["error", "never"]
  }
}
```

BPScott:
@robertoentringer for the most part you shouldn't ever need to set an explicit parser as it should be inferred by the filename, or indeed any other options (as they should be read from your .prettierrc per [this note](https://github.com/prettier/eslint-plugin-prettier#options).

What's the filename of the file that you're parsing that causes this to appear? A better fix would probably be to use the overrides section of your prettierrc file to specify the parser for the file extension you desire. [This comment](https://github.com/prettier/stylelint-prettier/issues/2#issuecomment-418883185) offers instructions on how you can set those the parser using overrides. (that comment is relating to stylelint-prettier but the concept of configuring prettier overrides in prettier's config files instead of eslint/stylelint's config is the same.

robertoentringer:
@BPScott  the filename is index.**html** and I use the **eslint-plugin-prettier**.

Setting the parser for html's files inside the **.prettierrc** file **works** very well: 
```json
{
  "overrides": [
    {
      "files": "*.html",
      "options": { "parser": "babel" }
    }
  ]
}
````
However, I like to use a **single file** for config **prettier+eslint**.

But, if I use the same config inside the  **.eslintrc.js** it has no any effect! Is there any way to implement this? I did not find anything about it at [options](https://prettier.io/docs/en/options.html).

```js
module.exports = {
  env: { browser: true, es6: true },
  extends: ["eslint:recommended", "plugin:prettier/recommended"],
  parserOptions: { ecmaVersion: 2018 },
  plugins: ["prettier", "html"],
  rules: {
    "prettier/prettier": [
      "error",
      {
        semi: false,
        printWidth: 100,
        overrides: [
          {
            files: "*.html",
            options: { parser: "babel" }
          }
        ]
      }
    ],
    "comma-dangle": ["error", "never"]
  }
}
```


This is my ***package.json*** : 

```json
{
  "scripts": {
    "lint": "eslint --fix src/**/*.{js,html}"
  },
  "devDependencies": {
    "eslint": "^5.16.0",
    "eslint-config-prettier": "^4.3.0",
    "eslint-plugin-html": "^5.0.5",
    "eslint-plugin-prettier": "^3.1.0",
    "prettier": "^1.18.2"
  }
}
```` 
Thank you for the good advice about [this note](https://github.com/prettier/eslint-plugin-prettier#options) and [this comment](https://github.com/prettier/stylelint-prettier/issues/2#issuecomment-418883185).

BPScott:
Setting overrides would work in every case but this - as per [the comment](https://github.com/prettier/eslint-plugin-prettier/blob/bca77e66ed1eba682eb13055862adc70478d472b/eslint-plugin-prettier.js#L189) above the line in your PR we force the parser value for html files because we want to try and automatically do the right thing in 90% of the cases rather than forcing people to add this sort of config.

It sounds like the fix is for us to update the parser name that we force based upon  `prettier.getSupportInfo()` as per my initial comment in this thread

robertoentringer:
@BPScott, based in your reply I tried to work  in that [solution](https://github.com/robertoentringer/eslint-plugin-prettier/commit/81f4a0053dbe0282be6500edfa584e52a9761c06).

Please, can you tell me what you think of this patch, before I send the PR?

BPScott:
That looks great !thanks for investigating the structure of that.

To save you some work I've added this in #212

PR #212, opened 2019-07-04
When forcing the JS parser, use the modern name

BPScott:
This avoids a deprecation notice

With thanks to @robertoentringer for spotting this

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -210,7 +210,13 @@ module.exports = {
             if (
               parserBlocklist.indexOf(prettierFileInfo.inferredParser) !== -1
             ) {
-              initialOptions.parser = 'babylon';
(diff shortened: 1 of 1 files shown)


Standard8:
@BPScott we've been hitting this on a project we're working. We have a workaround for it that we can easily use, but I was just wondering if you had any plans to release a new version soon?

BPScott:
Alas I've not got the keys to npm. @not-an-aardvark could you push out a 3.1.1 point release. (or setup publish rights if you want [me](https://www.npmjs.com/~bpscott) to be self sufficient)

not-an-aardvark:
@BPScott Added you to the package on npm. (Sorry, not sure if it sent out a notification email automatically.)

BPScott:
Didn't get an email but I've got access, thanks :)

v3.1.1 has just been published with this change.

PR #226, opened 2019-09-10
Added new options with default options for prettier

Tom910:
When developing for eslint plugins need set default options on prettier and add opportunity change default

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,16 @@ For the list of every available exclusion rule set, please see the [readme of es
       }]
       ```
 
+    - `defaultPrettierOptions`: An object representing [options](https://prettier.io/docs/en/options.html) that will be passed into prettier with lower priority and will be replaced options from `.prettierrc` and the first option
+
+      ```json
+      "prettier/prettier": ["error", {}, {
+        "defaultPrettierOptions": {
+          "singleQuote": true
+        }
+      }]
+      ```
+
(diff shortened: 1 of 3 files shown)


BPScott:
Hi @Tom910 what do you need this for? I'm having a hard time working out a reason why the existing prettier options object does not fit your needs.

I can't think of a reason why you'd want defaults that differ from the values you set in your project's `.prettierrc` file as that will lead to mismatches between when you run prettier standalone and via this plugin.

Tom910:
Hi, @BPScott  I have a library for eslint config, for example, I take  https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/rules/prettier.js now when I install the plugin I have required to setting prettier in my project and set basic configuration: singleQuote, maxWeight

But I don't want to require setting in the project. I want to work on the default setting and if the user wants to change settings, then he can create priitierrc and change default settings

BPScott:
Thanks for the info.

I'm strongly against adding any more features that may result in getting different results on if you run `prettier` or `eslint` (with the prettier plugin enabled). This feature allows for that to happen and thus increases the possibility for misconfiguration. You can achieve the same result by leveraging the standard prettier and eslint config files. Because of that I am not going to merge this feature.

Prettier recently added support for [extensible configs](https://prettier.io/docs/en/configuration.html#sharing-configurations). I suggest that you create and distribute a standard config for prettier in the same way that you distribute your eslint config. Then consuming projects can extend from that base prettier configuration.

Issue #184, opened 2019-04-18
Can't make usePrettierrc work under recommended config


maxiR8:
**What version of `eslint` are you using?**
eslint: "5.16.0

**What version of `prettier` are you using?**
prettier: 1.17.0

**What version of `eslint-plugin-prettier` are you using?**
eslint-plugin-prettier: 3.0.1

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
```
//.eslintrc
{
    "root": true,
    "env": {
        "browser": true,
        "es6": true,
        "commonjs": true,
        "jquery": true
    },
    "extends": ["plugin:prettier/recommended"],
    "globals": {
        "Atomics": "readonly",
        "SharedArrayBuffer": "readonly"
    },
    "parser": "babel-eslint",
    "rules": {
        "prettier/prettier": [
            "error",
            {},
            {
                "usePrettierrc": true
            }
        ]
    }
}
```
```
//.prettierrc.json

{
    "singleQuote": true,
    "trailingComma": "all",
    "semi": true,
    "tabWidth": 2
}
```

**What source code are you linting?**
Believe doesn't matter, any js code.

**What did you expect to happen?**
I am trying to setup eslint and prettier to work together under the same rules. I followed the instructions in README for Option 2 in using the usePrettierc, however, it works inconsistently when using the recommended config("extends": ["plugin:prettier/recommended"],). Expected proposed Option 2 in README (with usePrettierrc option) to work same as option 1.

**What actually happened?**
It doesn't follow the prettierrc options, or eslint complains that prettier formatted incorrectly (Ie regarding indentation.). This happens when using "extends": ["plugin:prettier/recommended"]. If I don't use the recommended config it works correctly.  I am using vscode prettier extension with "prettier.eslintIntegration": true. Maybe this is the cause.

Sorry If the mistake is on my end, a bit new to this.

BPScott:
`usePrettierrc: true` is the default behaviour, you don't need to opt into this explicitly.

Did you also install `eslint-config-prettier` per the 1st part of the recommended configuration? If you've installed `eslint-config-prettier` and `eslint-plugin-prettier` then the following should work:

```
{
    "root": true,
    "env": {
        "browser": true,
        "es6": true,
        "commonjs": true,
        "jquery": true
    },
    "extends": ["plugin:prettier/recommended"],
    "globals": {
        "Atomics": "readonly",
[...]

PR #240, opened 2019-10-30
Bump Prettier version to 1.18.2

skaldo:
Current version of prettier ignores (some) configuration options form .prettierrc.
With 1.18.2 all of the options are correctly applied.

Should fix https://github.com/prettier/eslint-plugin-prettier/issues/184 as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "eslint-plugin-node": "^8.0.0",
     "eslint-plugin-self": "^1.1.0",
     "mocha": "^6.0.0",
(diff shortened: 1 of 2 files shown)


BPScott:
This project shouldn't always need to be on the latest version of prettier - updating would mean we force projects to change their formatting to prettier's newest style which may be undesirable when they just want bugfixes.

You should be able to deduplicate the version of prettier in your project's package-lock.json / yarn.lock to ensure that you have one version of prettier installed in your dependencies without needing us to increase the version.

PR #241, opened 2019-10-30
fix: add eslint-config-prettier to dependencies

sastan:
If this plugin is used with pnpm eslint fails because it can not find `eslint-config-prettier` in the non-flat `node_modules`.

This PR adds `eslint-config-prettier` to the `dependencies`.

Another solution would be to add `eslint-config-prettier` to the `peerDependencies`. I would change the PR if that would be the better choice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   },
   "homepage": "https://github.com/prettier/eslint-plugin-prettier#readme",
   "dependencies": {
+    "eslint-config-prettier": "^6.0.0",
     "prettier-linter-helpers": "^1.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


BPScott:
It is possible to use this plugin without relying on eslint-config-prettier - in particular if you wish to manage the list of disabled rules yourself. We do not directly depend upon that project within our own codebase, aside from the recommended config whose that usage is optional.

By not adding this dependency (or a peerDependency) we ensure that consumers can provide their desired version of eslint-config-prettier instead of us having to keep up to date with latest version.

If you add `eslint-config-prettier` to your consuming project (per [the directions](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) for the recommended config) does your project work?

sastan:
That makes sense. I found another way. Thank you.

PR #245, opened 2019-11-11
Build: update package.json and changelog for v3.1.2

sangeeth96:
The 1.19.1 release of Prettier has quite a lot of new things including TS 3.7 support. With the existing release, eslint and prettier seems to be odd at a few places. Here's an example:

image

Hope we can proceed with this release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## v3.1.2 (2019-11-11)
+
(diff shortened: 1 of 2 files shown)


BPScott:
The only thing that has changed since 3.1.1 is the yarn.lock file, which is specific to this project and has zero bearing on any project that consumes this package. You're free to update the lock file in your own applications to pull in the latest version of Prettier without needing a release of this project.

PR #283, opened 2020-04-11
Set `meta.type` to "layout"

brettz9:
Per https://eslint.org/docs/user-guide/command-line-interface#fix-type , one can set `meta` to "problem" or "suggestion" for code-related changes or "layout" for formatting ones. Setting this value allows `eslint --fix-type` to take into account the type of fix desired. (And I'm making an eslint badge formatter for which it is more convenient to have this meta info so any rule violations can be grouped by type.)


BPScott:
Thanks! Published in v3.1.3

brettz9:
Thank you!

PR #288, opened 2020-04-22
Prettier 2

brettz9:
Builds on #287, with updates to get builds passing.

- Breaking change: Update `engines` to new prettier Node minimum (10.13.0)
- Travis: update/simplify travis.yml according to new minimum (adding Node 12 as well)
- npm: Updated prettier peerDep. Also updated eslint peerDep. (new minimum version of Node (10.13) for prettier makes eslint 5 no longer necessary to support)
- npm: Updated available devDeps.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,12 +6,12 @@ module.exports = {
     'not-an-aardvark/node',
     'plugin:node/recommended',
     'plugin:eslint-plugin/recommended',
-    'prettier'
+    'prettier',
   ],
   env: { mocha: true },
   root: true,
   rules: {
     'self/prettier': ['error'],
-    'eslint-plugin/report-message-format': ['error', '^[^a-z].*\\.$']
-  }
+    'eslint-plugin/report-message-format': ['error', '^[^a-z].*\\.$'],
+  },
 };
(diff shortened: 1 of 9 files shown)


brettz9:
I've added some commits so it is now in fact passing.

BPScott:
Does dropping support for prettier 1.x, eslint 5.x and node 8.x  actually unblock any kind of new feature or allow any existing code to be considerably simplifie? I know there's a few conditional checks for recent-ish prettier API features but they're small-fry.

Consumers are able already to use prettier v2 with this project so we're not unlocking any new functionality there.

From what I can tell this PR requires a major version bump just for the sake of updating a development dependency. It seems to be churn for the sake of churn right now. Eventually we'll want to make this change it but unless there is a positive upside to make this change worth doing now I'd like to give a decent amount of time for folks to move away from EOL versions before making this change.

brettz9:
> Does dropping support for prettier 1.x, eslint 5.x and node 8.x actually unblock any kind of new feature or allow any existing code to be considerably simplifie? I know there's a few conditional checks for recent-ish prettier API features but they're small-fry.
> 
> Consumers are able already to use prettier v2 with this project so we're not unlocking any new functionality there.

I could have sworn I was having trouble getting it to use v2, which is why I prepared this, but in re-checking now, it does seem to be reinforcing v2, so will close. Sorry for the distraction...

PR #335, opened 2020-09-12
docs(README): fix prettier getFileInfo link

neighborhood999:
Fix prettier `getFileInfo` link.


BPScott:
Thanks for spotting that got updated!

Issue #65, opened 2017-10-16
prefer-arrow-callback fix conflicting with prettier fix


ismail-syed:
**What version of `eslint` are you using?**
`v4.9.0`

**What version of `prettier` are you using?**
`v1.7.4`

**What version of `eslint-plugin-prettier` are you using?**
`v2.3.1`

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
https://github.com/ismail-syed/prettier-eslint-config-invalid-code

**What source code are you linting?**
```
function foo() {
  return isTrue && [0,1,2].map(function(num) {
    return num * 2;
  });
}
```

**What did you expect to happen?**
The code above should be formatted as per prettiers config and also should adhere to that `prefer-arrow-callback` fix

**What actually happened?**
Invalid code was generated, closing parenthesis is missing on the return statement.
```
function foo() {
  return (
    isTrue &&
    [0, 1, 2].map((num) => {
    return num * 2;
  });
}

```

Is the underlying issue from the `prefer-arrow-callback` fixer or the prettier plugin fixer?

lydell:
> Is the underlying issue from the `prefer-arrow-callback` fixer or the prettier plugin fixer?

It could also be how ESLint applies fixes.

not-an-aardvark:
Related: https://github.com/eslint/eslint/issues/7928

tl;dr: Autofixes from rules can, on rare occasions, conflict with each other. Rules can work around the issue by coarsening their fixes to ensure that the "fix range" contains the entire range of text that caused the particular problem to be reported, rather than just the actual text replacement.

For convenience, could you paste the output of running ESLint without `--fix` when those two rules are enabled? I think this might be a problem with the core `prefer-arrow-callback` rule, but I'm not sure.

ismail-syed:
Here is the output of running `ESLint` without `--fix` on `issue2.js`. 
```
╰─$ yarn run lint
yarn run v1.1.0
$ eslint issue2.js --max-warnings 0

/Users/ismailsyed/test/prettier-eslint-config-invalid-code/issue2.js
  2:10  error  Replace `isTrue·&&·[0,1,` with `(⏎····isTrue·&&⏎····[0,·1,·`  prettier/prettier
  2:32  error  Unexpected function expression                                prefer-arrow-callback
  3:5   error  Insert `··`                                                   prettier/prettier  4:1   error  Replace `··}` with `····})⏎··`                                prettier/prettier

[...]

PR #360, opened 2020-11-29
Turn off problematic rules in recommended config (prepare for next eslint-config-prettier version)

lydell:
https://github.com/prettier/eslint-config-prettier/pull/168 is moving `arrow-body-style` and `prefer-arrow-callback` (which conflict with eslint-plugin-prettier, #65) from the `prettier` config to its own file, `prettier/prettier`. The reason for doing this is:

- We don’t want to turn off those rules for all Prettier users who _don’t_ use eslint-plugin-prettier.
- Since those exclusions are specific to eslint-plugin-prettier, it makes sense to put them in a separate file just like we do for all other plugins.

This PR turns off those two rules in the `recommended` config. This helps people upgrade:

- People can upgrade eslint-plugin-prettier but not eslint-config-prettier without any problems.
- People who upgrade both can keep using the `recommended` config without issues.

This PR also moves documentation about all of this from eslint-config-prettier to this repo.

This can and should be released before https://github.com/prettier/eslint-config-prettier/pull/168

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,6 @@ This plugin ships with a `plugin:prettier/recommended` config that sets up both 
    npm install --save-dev eslint-config-prettier
    ```
 
-   Compatibility:
-
-   - eslint-plugin-prettier@>=4 ↔ eslint-config-prettier@>=7
-   - eslint-plugin-prettier@<4 ↔ eslint-config-prettier@<7
-
 2. Then you need to add `plugin:prettier/recommended` as the _last_ extension in your `.eslintrc.json`:
 
    ```json
@@ -76,8 +71,6 @@ This plugin ships with a `plugin:prettier/recommended` config that sets up both 
      "extends": ["plugin:prettier/recommended"]
    }
    ```
-
-   See [eslint-config-prettier’s explanation](https://github.com/prettier/eslint-config-prettier#eslint-plugin-prettier) for more information about what this config does.
 
    You can then set Prettier's own options inside a `.prettierrc` file.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -52,48 +54,52 @@ Then, in your `.eslintrc.json`:
...
+2. Then you need to add `plugin:prettier/recommended` as the _last_ extension in your `.eslintrc.json`:
 
    ```json
    {
      "extends": ["plugin:prettier/recommended"]
    }
    ```
 
-This does three things:
+   See [eslint-config-prettier’s explanation](https://github.com/prettier/eslint-config-prettier#eslint-plugin-prettier) for more information about what this config does.
```
lydell, line 80: This link is broken at the moment but will work once https://github.com/prettier/eslint-config-prettier/pull/168 is merged.

Preview: https://github.com/prettier/eslint-config-prettier/tree/next#eslint-plugin-prettier

BPScott:
Thanks @lydell, those readme changes look great!

As you've talked about in https://github.com/prettier/eslint-config-prettier/tree/next#eslint-plugin-prettier the relationship between these two packages has the potential to be real confusing. I don't think bouncing people between docs for the two repos helps that.

It feels like explaining eslint-plugin's shortcomings and conflicts within eslint-plugin-prettier's README would make more sense than having to do it in eslint-config-prettier's README (and then linking to that from eslint-plugin-prettier's readme anyway!)

What if instead of adding the ignoring of `arrow-body-style and prefer-arrow-callback` to a new config in eslint-config-prettier, we instead added then to eslint-plugin-prettier's "recommended" config and [all the docs about how eslint-plugin-prettier doesn't play nicely with core eslint rules](https://github.com/prettier/eslint-config-prettier/tree/next#arrow-body-style-and-prefer-arrow-callback) could live in this repo instead?

So our recommend config would be:

```
recommended: {
      extends: ['prettier'],
      extends: ['prettier'],
      plugins: ['prettier'],
      rules: {
        'prettier/prettier': 'error',
        'arrow-body-style': 0,
        'prefer-arrow-callback': 0,
      }
}
```

That'd mean eslint-plugin-prettier would be responsible for disabling its own conflicting rules, and it wouldn't require a major version bump in eslint-plugin-prettier.

lydell:
Great idea! I’ll update soon.

lydell:
I’ve updated according to your suggestions and just re-read the readme – it feels good. Ready for another round of review! The idea is to merge this before releasing v7 of eslint-config-prettier.

BPScott (approved):

BPScott:
This looks great, thanks for the updates!

I'll merge this, stop testing CI on node 8 so builds get back to passing, then cut a release

BPScott:
There we go. just published as v3.2.0

lydell:
Thanks! 🎉

PR #367, opened 2020-12-13
Add peerDependenciesMeta block

BPScott:
Fixes warnings when using Yarn v2

Fixes #322

PR #374, opened 2021-01-04
fix: add eslint-config-prettier as an optional peer dependency

merceyz:
**What's the problem this PR addresses?**

https://github.com/prettier/eslint-plugin-prettier/pull/367 added a peer dependency on `eslint-plugin-prettier` instead of `eslint-config-prettier`

**How did you fix it?**

Changed the peer dependency from `eslint-plugin-prettier` to `eslint-config-prettier`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "vue-eslint-parser": "^6.0.0"
   },
   "peerDependenciesMeta": {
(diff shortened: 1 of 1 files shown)


BPScott:
I am an idiot! Excellent spot, thanks @merceyz!

PR #379, opened 2021-01-22
Fix conflicts with `arrow-body-style`

fisker:


fisker:
I come up with a better solution, will give another shot.

PR #380, opened 2021-01-23
Simplify report logic

fisker:
The old one is too complicated.


3124o (approved):

BPScott (approved):
This is so much nicer! Thanks @fisker

fisker:
That's the plan.

PR #417, opened 2021-04-29
New: adds option to prioritize merging of options

### LINKED ISSUE #416: Extending Default Config in a Package
I want to share a default config which is easily overridable by a project's prettier config.

I expect to be able to configure default prettier settings in eslint, which are overridden by a project's defined prettier settings.

I see two scenarios:

1. **A project has a .prettierrc file**, and they adopt our shared eslint config: their .prettierrc file should take precedence.  If they want to adopt the shared prettier config, they can simply delete their .prettierrc file.
2. **A project does not have a .prettierrc file**, and they adopt our shared eslint config: our shared prettier config should become the new defaults.  If they want to make changes, they can simply create a .prettierrc file.

**What actually happened?**

**I can't** set ESLint prettier settings because that will override local prettier config.

**I can't _not_** set ESLint prettier settings because there appears to be no other way to distribute shared default prettier (current interim solution is to provide a shared .prettierrc file for a local .prettierrc file to extend from, but this makes for the annoying creation of a .prettierrc.js file just to require the package's .prettierrc file)

**Main Question**

Can we just alter the priority of the prettier options?

https://github.com/prettier/eslint-plugin-prettier/blob/master/eslint-plugin-prettier.js#L223-L229
```js
const prettierOptions = Object.assign(
  {},
  initialOptions,
  prettierRcOptions,
  eslintPrettierOptions,
  { filepath }
);
```

Why aren't the prettierRcOptions and eslintPrettierOptions swapped?  Maybe the plugin could provide an option as to which options should take precedence?

bozdoz:
Closes #416 

By creating a new option to configure the merging of prettier options, our shared eslint config package can now defer to a consuming project's `.prettierrc` file (or lack thereof).  With this option, our project can set default rules only in eslint, and each consuming project can easily extend those default rules.

Basic Example:

```js
// DRY - import options from base
const prettier = require('./.prettierrc.json');

module.exports = {
    rules: {
        'prettier/prettier': ['error', prettier, {
          priorityOptions: 'prettierRc',
        }],
    },
};
```

Now eslint has the prettier options, I've de-duped the options by simply requiring the package's prettierrc.json file, and consuming projects can install this package with two scenarios:
1. project **has** a .prettierrc file: newly installed shared eslint config defers to the local .prettierrc options and the consumer can remove the .prettierrc file, or remove unnecessary options.
2. project **does not have** a .prettierrc file: newly installed shared eslint config provides default options that differ from prettier defaults.

This makes sharing and extending much easier.  
1. **No need** to alter `.eslintrc` to override prettier settings there.
2. **No need** to have a `.prettierrc` file at all.
3. **No need** to extend using a `.prettierrc.js` file.

Any comments welcome!  Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ If you’re fixing large of amounts of previously unformatted code, consider tem
     "prettier/prettier": ["error", {"singleQuote": true, "parser": "flow"}]
     ```
 
-    NB: This option will merge and override any config set with `.prettierrc` files
+    NB: This option will merge and override any config set with `.prettierrc` files (unless you specify `priorityOptions: 'prettierRc'` in the second option (see below))
 
 - The second option:
 
@@ -138,6 +138,19 @@ If you’re fixing large of amounts of previously unformatted code, consider tem
         "fileInfoOptions": {
           "withNodeModules": true
         }
+      }]
+      ```
+
+    - `priorityOptions`: An enum deciding which options should take precedence when merging. You might want to set `prettierRc` as the priority if you want to use `.prettierrc` files to override eslint settings (useful for distributing a shared eslint config).
+
+      - `prettierRc`: the `.prettierrc` file options take precedence over the eslint prettier options
+      - `eslint` (default): as stated above, eslint options will merge and override any config set with `.prettierrc` files
+
+      ```json
+      "prettier/prettier": ["error", {
+        tabWidth: 4
+      }, {
+        "priorityOptions": "prettierRc"
       }]
       ```
 
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -122,6 +122,11 @@ module.exports = {
                 type: 'object',
                 properties: {},
                 additionalProperties: true
+              },
+              priorityOptions: {
+                type: 'string',
+                enum: ['prettierRc', 'eslint'],
+                default: 'eslint'
               }
             },
             additionalProperties: true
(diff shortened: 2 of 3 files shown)


buschtoens:
I would *love* to see this getting merged! Is there anything holding it back?

BPScott:
Heya folks, Thank you for the PR but I'm don't think I'm going to merge this. Being able to adjust configuration ordering feels like a big footgun, as it opens the door to confusing and inconsistent behaviour when you try and mix config in prettierrc and eslint. I consider the current ability to override prettier's options in your eslint file a problematic historical relic, and I don't want to encourage its future use.

We already read the contents of the `.prettierrc` file and use it, so you should try and keep all prettier config in that file rather than placing your prettier config somewhere where the `prettier` executable can not find it. 

This is because by allowing the configuration of prettier inside eslint config files you then run into inconsistencies when you run `eslint` or `prettier` over a given file - `eslint` formats files one way, `prettier` formats another as the options passed to prettier differ. This can also effect files that embed other languages - for instance javascript in `.js` files will use the prettier config from your eslint config, but javascript in a `js` fenced code block in a `.md` would not use the prettier config from your eslint config. This fighting is particularly problematic if your editor is setup to run prettier and eslint when saving files.

This linter fighting has the potential to be confusing for end-users so I don't want to encourage scenarios where it can happen.

---

I would recommend that instead of specifying prettier defaults inside your shared eslint config, you create [a shared package that contains your prettier config](https://prettier.io/docs/en/configuration.html#sharing-configurations) and tell end-users to extend from it if they wish to use prettier:

```
// your shareable eslint config published as @mycompany/eslint-plugin
rules: {
  'prettier/prettier': ['error'],
},


// your shareable prettier config published as @my-company/prettier-config
{
  //whatever
}

// In a consuming projects's eslint config or the `eslint` key of its package.json
{extends: '@mycompany/eslint-plugin'}

// In a consuming project's prettierrc, or in the `prettier` key of it's package.json
"@my-company/prettier-config"

PR #429, opened 2021-08-23
v4 - Drop support for eslint 5/6, prettier 1, node 6/8

BPScott:
This PR contains breaking changes that shall go into v4 of eslint-plugin-prettier. API and usage remains the same, however we drop support for old and unsupported versions of eslint, prettier and node. This allows us to use recent features like eslint's `context.getPhysicalFilename()` instead of writing workarounds.

- eslint v7.28.0 is the minimum supported version of eslint
- prettier v2.0.0 is the minimum supported version of prettier
- node v12 is the minimum supported version of node (we don't enforce this with an engines field, and this should work on node 10 but we're not going to test this on CI so use the now-EOL node 10 at your own risk)

### DIFF
--- a/test/prettier.js
+++ b/test/prettier.js
@@ -166,7 +166,7 @@ function loadInvalidFixture(name) {
   const src = fs.readFileSync(filename, 'utf8');
   const sections = src
     .split(/^[A-Z]+:\n/m)
-    .map(x => x.replace(/(?=\n)\n$/, ''));
+    .map((x) => x.replace(/(?=\n)\n$/, ''));
   const item = {
     code: sections[1],
     output: sections[2],
--- a/test/prettier.js
+++ b/test/prettier.js
@@ -44,23 +44,23 @@ ruleTester.run('prettier', rule, {
     {
       code: `var foo = {bar: 0};\n`,
       filename: getPrettierRcJsFilename('bracket-spacing'),
-      options: [{ bracketSpacing: false }]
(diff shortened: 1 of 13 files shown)


test/prettier.js
```diff
@@ -124,29 +87,71 @@ ruleTester.run('prettier', rule, {
...
+// https://github.com/apollographql/eslint-plugin-graphql/blob/c465fedc8fea239ee1731ad4ec3ee1183a3cdddf/src/index.js#L404
+// In the future if ESLint supports processors (https://github.com/eslint/rfcs/pull/31)
+// we should be define a RuleTester like
+// `newRuleTester({processor: require('eslint-plugin-graphql').processor['.graphql']})
+// and then pass in pure graphql into the code value.
+const eslintPluginGraphqlRuleTester = new RuleTester({
+  parserOptions: { ecmaVersion: 2015 },
+});
+
+eslintPluginGraphqlRuleTester.run('eslint-plugin-graphql', rule, {
```
MichaelDeBoey, line 147: Shouldn't `eslint-plugin-graphql` be a dependency too? 🤔

BPScott: No. This test does not run eslint-plugin-graphql currently, it merely simulates what that plugin's processors would output. As there it is not possible to run the processors from that plugin per the big comment above this line.

MichaelDeBoey (approved):

PR #467, opened 2022-04-19
Bump prettier to 2.6.2

rbargholz:
The latest prettier verison treats Angular's waitForAsync like JS's native `async` operator. Having prettier and ESLint running `eslint-plugin-prettier`both trying to format a file causes a conflict.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-plugin-self": "^1.2.1",
     "graphql": "^15.7.1",
     "mocha": "^6.2.3",
-    "prettier": "^2.4.1",
+    "prettier": "^2.6.2",
     "vue-eslint-parser": "^8.0.0"
   },
   "peerDependenciesMeta": {


rbargholz:
@BPScott friendly ping on this one. Wondering if you can give some insight into whether bumping prettier is a harmless chore for this library, or whether it needs more consideration.

BPScott:
This PR is currently only changing a devDependency - it has zero effect on what packages get installed by products that use eslint-plugin-prettier.


Increasing the required prettier version that consumers would be a change to decrease the allowed options in the peerDependencies which is technically a breaking change, as we would be saying "something that worked ok up till now will now cause errors".

Can you provide an example of what this conflict is? You shouldn't enable eslint rules that conflict with prettier. This is because prettier's ethos is that [prettier should handle code formatting, while linters should handle code quality](https://prettier.io/docs/en/comparison.html). I wonder if you need to update your version of  https://github.com/prettier/eslint-config-prettier to disable all rules that would conflict with prettier. That package also provides a helper tool that lets you identify any rules that you currently have enabled that would conflict with prettier's styling.

rbargholz:
@BPScott Thanks for the response. Very sorry for not being so clear. I'll give an example.

It's not a conflict between ESLint rules and prettier, I also agree that ESLint should handle code quality. Rather `prettier@2.6.2` treats Angular's `waitForAsync` keyword like the native JS `async` keyword. I faced a conflict in April between my VSCode ESLint extension whereby `eslint-plugin-prettier` with `prettier@2.4.1` and `prettier@2.6.2` installed in the project would undo each other's formatting when running `npx eslint --fix` on the CLI and then formatting on save later. I assumed this was due to `eslint-plugin-prettier` running `prettier@2.4.1` as an ESLint rule (Not in a project-specific `npx` context) which is lower than the version of `prettier` we have installed in the project (Running in project-specific `npx` context). You'd be the best person to know if this is an issue with `eslint-plugin-prettier` or not, though.

BPScott:
`eslint-plugin-prettier` doesn't have a direct dependency on prettier at all - it has a peerDependency, which means it picks up what ever you specify in your package.json.

That sounds like you did an `npm install` to update a version of prettier in your project and then didn't restart VSCode - which would use the a cached version of the older prettier version until VSCode's eslint bits were restarted (the easiest way to do this being "Restart VSCode").

Having two versions of prettier installed in your project seems incorrect - you should be be to deduplicate that down into a single version.

That seems unrelated to any setup within `eslint-plugin-prettier` though. With that in mind, I'm going to close this.

PR #477, opened 2022-06-29
chore: use changesets for version management

JounQin:


.github/workflows/pkg-size.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+          cache: yarn --frozen-lockfile
```
JounQin, line 21:
```suggestion
          cache: yarn
```

PR #504, opened 2022-08-12
Remove `Sponsors` and `Backers` section from README.md

fisker:
For the fundings, I suggest you open a joined account on opencollective, but before that I guess we can use @JounQin's personal account.

But I don't think display all those unrelated orgs in README is the right thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,22 +173,6 @@ If you’re fixing large of amounts of previously unformatted code, consider tem
 
 ---
 
-## Sponsors
-
-| 1stG                                                                                                                               | RxTS                                                                                                                               | UnTS                                                                                                                               |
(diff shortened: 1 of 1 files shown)


JounQin:
Sure

JounQin:
@fisker For opencollective, how can I `join prettier`?

fisker:
I  mean  a new account, like  other projects did. https://github.com/prettier/plugin-ruby/blob/main/.github/FUNDING.yml

JounQin:
> I mean a new account, like other projects did. [https://github.com/prettier/plugin-ruby/blob/main/.github/FUNDING.yml](https://github.com/prettier/plugin-ruby/blob/main/.github/FUNDING.yml?rgh-link-date=2022-08-13T01%3A21%3A04Z)

OK, that seems to make sense. I'll do it later.

PR #505, opened 2022-08-12
chore: remove personal config

fisker:
I don't see why we need those package of configs.

If you want introduce new plugins/rules, just add them.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,32 +4,15 @@
     "mocha": true
   },
   "extends": [
-    "@1stg",
-    "plugin:eslint-plugin/recommended"
+    "plugin:n/recommended",
+    "plugin:sonar/recommended",
+    "plugin:sonarjs/recommended",
+    "plugin:eslint-plugin/recommended",
+    "prettier"
   ],
+  "plugins": ["self"],
   "rules": {
-    "eslint-plugin/report-message-format": [
-      "error",
-      "^[^a-z].*\\.$"
-    ]
-  },
-  "overrides": [
-    {
-      "files": "*.js",
-      "plugins": [
-        "self"
-      ],
-      "rules": {
-        "self/prettier": [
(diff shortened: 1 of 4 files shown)


JounQin:
Because it is easier to maintain access projects, we do have personal configs previously, but those seems unmaintained and very outdated. But I'm fine to be verbose here.

.eslintrc.json
```diff
@@ -4,32 +4,15 @@
     "mocha": true
   },
   "extends": [
-    "@1stg",
-    "plugin:eslint-plugin/recommended"
```
JounQin, line 8: A lot of plugins are missing in this PR.

JounQin (changes requested):

.eslintrc.json
JounQin: And also there some special logic inside

1. https://github.com/1stG/configs/blob/master/packages/config/index.js
2. https://github.com/1stG/configs/blob/master/packages/lint-staged/base.js#L7-L8

.prettierrc
```diff
@@ -1 +1,5 @@
-"@1stg/prettier-config/semi"
+{
```
JounQin, line 1: https://github.com/1stG/configs/blob/master/packages/prettier-config/base.js#L22-L48

JounQin (changes requested):

JounQin:
Close because a lot of features are missing.

.eslintrc.json
fisker: It's hundreds line project! What for ?

.prettierrc
fisker: Same as above.

.eslintrc.json
JounQin: That's why I'm using presets here.

PR #561, opened 2023-06-20
feat: introduce option to force format extracted fragements

paescuj:
The default behavior skipping the formatting of extracted fragments (e.g. code blocks in Markdown files, parsed via `eslint-plugin-markdown`) is usually fine. https://github.com/prettier/eslint-plugin-prettier/blob/00449dfd152dbdd5a0a52426bb538dddcf702328/worker.js#L127-L150

However, there are cases where you want to have such fragments formatted anyways.
This pull request introduces the new option `forceFormatExtracted` (happy to change the naming) which turns this particular check off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,22 @@ If you’re fixing large of amounts of previously unformatted code, consider tem
       }
       ```
 
+    - `forceFormatExtracted`: Enables formatting of extracted fragements, (default: `false`).
+      ESLint supports processors that let you extract and lint JS fragments within a non-JS language, for example using `eslint-plugin-markdown` to extract and lint a code block inside a Markdown file. Usually, you don't want to have the fragement itself formatted by Prettier as the whole file
+      might already be formatted anyway. This is why `eslint-plugin-prettier` ignores such fragements by default. However, for cases you want to have such extracted fragements formatted by Prettier you can use this option to force it.
+
+      ```json
+      {
+        "prettier/prettier": [
+          "error",
+          {},
+          {
+            "forceFormatExtracted": true
+          }
+        ]
+      }
+      ```
+
     - `fileInfoOptions`: Options that are passed to [prettier.getFileInfo](https://prettier.io/docs/en/api.html#prettiergetfileinfofilepath--options) to decide whether a file needs to be formatted. Can be used for example to opt-out from ignoring files located in `node_modules` directories.
 
       ```json
--- a/.changeset/strong-lions-hang.md
+++ b/.changeset/strong-lions-hang.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-prettier": minor
+---
+
+Introduce option to force format extracted fragements
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
+.changeset/
 # this file doesn't exist, but we use it as a filename that should be ignored
 # by prettier in the tests
 ignore-me.js
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -10,7 +10,7 @@
  * @typedef {import('eslint').AST.SourceLocation} SourceLocation
  * @typedef {import('eslint').ESLint.Plugin} Plugin
  * @typedef {import('prettier').FileInfoOptions} FileInfoOptions
- * @typedef {import('prettier').Options & { onDiskFilepath: string, parserPath: string, usePrettierrc?: boolean }} Options
+ * @typedef {import('prettier').Options & { onDiskFilepath: string, parserPath: string, usePrettierrc?: boolean, forceFormatExtracted?: boolean  }} Options
  */
 
 'use strict';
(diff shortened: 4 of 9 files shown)


README.md
```diff
@@ -153,6 +153,22 @@ If you’re fixing large of amounts of previously unformatted code, consider tem
       }
       ```
 
+    - `forceFormatExtracted`: Enables formatting of extracted fragements, (default: `false`).
+      ESLint supports processors that let you extract and lint JS fragments within a non-JS language, for example using `eslint-plugin-markdown` to extract and lint a code block inside a Markdown file. Usually, you don't want to have the fragement itself formatted by Prettier as the whole file
+      might already be formatted anyway. This is why `eslint-plugin-prettier` ignores such fragements by default. However, for cases you want to have such extracted fragements formatted by Prettier you can use this option to force it.
```
JounQin, line 158: > However, for cases you want to have such extracted fragements formatted by Prettier you can use this option to force it.

So, the case is, you want to format js in markdown via prettier, but not the markdown file itself? cc @BPScott

JounQin (changes requested):

README.md
paescuj: Yeah, sort of. Actually I'd like to be able to apply a different set of Prettier rules to code blocks in Markdown files resp. be able to individually handle those blocks (for example disabling Prettier for JS code blocks, while formatting other code blocks like JSON). This can be achieved via ESLint rules on Markdown files processed with `eslint-plugin-markdown`. This set-up is working fine, but only if I'm able to disable the default behavior of `eslint-plugin-prettier` with this newly introduced option to not skip those code blocks.

BPScott:
> However, there are cases where you want to have such fragments formatted anyways.

I was about to ask for an exact example of this case. Fortunately you've provided one in a comment already:

> Yeah, sort of. Actually I'd like to be able to apply a different set of Prettier rules to code blocks in Markdown files resp. be able to individually handle those blocks (for example disabling Prettier for JS code blocks, while formatting other code blocks like JSON). This can be achieved via ESLint rules on Markdown files processed with `eslint-plugin-markdown`. This set-up is working fine, but only if I'm able to disable the default behavior of `eslint-plugin-prettier` with this newly introduced option to not skip those code blocks.

---

This change feels like a really really big footgun. A big aim of `eslint-plugin-prettier` is that you should get non-conflicting results regardless of if you format a file through `prettier` or `eslint` with `eslint-plugin-prettier` enabled. In order to do this, we must leverage existing eslint/prettier configuration - usually through overrides - as much as possible instead of adding our own.

When we diverge we introduce the support burden of having to answer a bunch of support questions asking why formatting a file with prettier does one thing but eslint does another. This gets extra fun inside editors if you've got eslint and prettier plugins installed at the same time because now we have to explain potential for editor misconfiguration too.

`eslint-plugin-prettier` should strive to be glue between two other things that are configurable. Every option that we add (including the two existing `usePrettierrc`, and `fileInfoOptions` options) is a step away from that aim of `prettier` and `eslint` producing non-conflicting results, and thus I default to being against config option additions without a very very very good reason. In an ideal world I'd like to nuke those two existing options too, but I don't want to have to deal with the fallout of that either - we're in a nice place where the codebase is feature-complete and stable and doesn't need to be tinkered with for he most part.

If you're trying to do something particularly funky that truly can not be resolved by leveraging existing config, honestly I'd rather you fork `eslint-plugin-prettier` for your specific use-case instead of having to deal with the maintenance burden of supporting that weird use-case as I think additional options are ripe for misunderstanding.

---

I suspect we might be able to get your desired state another way without adding a config option.


`eslint-plugin-markdown` uses processors to generate custom virtual filenames for each of the fenced code blocks it finds in a markdown file, per [its readme](https://github.com/eslint/eslint-plugin-markdown#frequently-disabled-rules) it sounds like they're in the format `the-source-file.md/0.js` - with 0 increasing for each fenced codeblock it finds in that source file.


You should be able to add the following override to your eslint config to disable prettier in those js blocks within markdown files:

This would mean that those blocks would be not `prettier`ed when you run the file through eslint, but if you ever ran the file through `prettier` then you get a different result.

```js
 overrides: [
  {
    // 1. Target ```js code blocks in .md files.
    files: ["**/*.md/*.js"],
    rules: {
      "prettier/prettier": "off",
    }
  }
]
```

---

Alternatively you can prefix the fenced codeblocks that you don't want to format with prettier with a ignore comment to get prettier to skip it:

    # here is my markdown file
       
    Here is some text
    
    
    ```js
    ugly   ( codeThatIsIgnoredByPrettier ) ;
    ```
    
    Here is some more text


This is almost certainly the most idiomatic approach, which would result in the same behaviour when you run the file through both `prettier` and `eslint`.
You could probably write some kind of markdown linter that says "if you find a fenced codeblock with the `js` language, then it must be immediately preceded by a `<!-- prettier-ignore -->` comment.

README.md
BPScott: Commented down below. I think there's ways of doing this without introducing a new option.

JounQin:
close as above comment

Issue #563, opened 2023-07-05
[prettier v3]: TypeError: Cannot read properties of null (reading 'map')


auvred:
**What version of `eslint` are you using?**
8.44.0

**What version of `prettier` are you using?**
3.0.0

**What version of `eslint-plugin-prettier` are you using?**
5.0.0-alpha.1

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
```jsonc
// .eslintrc
{
  "plugins": ["prettier"],
  "rules": {
    "prettier/prettier": ["error"]
  }
}
```

**What source code are you linting?**
Any code, for example `console.log(1)`

**What did you expect to happen?**
Lint without errors

**What actually happened?**
```
$ pnpm eslint . 

Oops! Something went wrong! :(

ESLint: 8.44.0

TypeError: Cannot read properties of null (reading 'map')
    at loadPlugins (file:///my-project/node_modules/.pnpm/prettier@3.0.0/node_modules/prettier/index.mjs:21518:30)
    at Module.<anonymous> (file:///my-project/node_modules/.pnpm/prettier@3.0.0/node_modules/prettier/index.mjs:21781:9)
    at /my-project/node_modules/.pnpm/eslint-plugin-prettier@5.0.0-alpha.1_eslint@8.44.0_prettier@3.0.0/node_modules/eslint-plugin-prettier/worker.js:44:56
```

PR #564, opened 2023-07-05
fix: compatibility with prettier@3 without plugins

auvred:
Fixes https://github.com/prettier/eslint-plugin-prettier/issues/563

In `prettier@3.0.0` the behavior of plugins list default value was changed

from this:

https://github.com/prettier/prettier/blob/e154250e19caa6d1b23bb1e7803ee5f8d0f95445/src/common/load-plugins.js#L25-L27

to this:

https://github.com/prettier/prettier/blob/266344398c1c90f7dc7b206ba8077cecc294470f/src/main/plugins/load-plugins.js#L3

So `plugins: null` causes runtime error, the default value should be `undefined` as declared [there](https://github.com/prettier/prettier/blob/266344398c1c90f7dc7b206ba8077cecc294470f/src/index.d.ts#L784)

### DIFF
--- a/.changeset/stale-guests-battle.md
+++ b/.changeset/stale-guests-battle.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-prettier": patch
+---
+
+fix: compatibility with prettier@3 without plugins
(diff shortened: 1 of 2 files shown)


Pnlvfx:
I can confirm that this fix the issue

PR #568, opened 2023-07-11
chore: bump deps, use pnpm instead

JounQin:


BPScott:
@JounQin thank you for dealing with all the async->sync worker threads stuff and the release of v5!

Issue #584, opened 2023-09-28
Issue parsing `.svelte` files when using Flat Config


eagerestwolf:
**What version of `eslint` are you using?**: `8.50.0`

**What version of `prettier` are you using?**: `3.0.3`

**What version of `eslint-plugin-prettier` are you using?**: `5.0.0`

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**

**`prettier.config.js`**
```js
/** @type {import("prettier").Config} */
export default {
  arrowParens: "always",
  bracketSameLine: true,
  bracketSpacing: true,
  endOfLine: "lf",
  htmlWhitespaceSensitivity: "css",
  overrides: [
    {
      files: "*svelte",
      options: {
        parser: "svelte"
      }
    }
  ],
  plugins: ["prettier-plugin-svelte", "prettier-plugin-tailwindcss"],
  printWidth: 100,
  proseWrap: "preserve",
  quoteProps: "as-needed",
  semi: true,
  singleAttributePerLine: false,
  singleQuote: false,
  svelteStrictMode: true,
  tabWidth: 2,
  trailingComma: "none",
  useTabs: false
};
```

**`eslint.config.js`**

For the sake of brevity, I didn't include all of my local includes and rules. This is an abbreviated version of the shared config I am using. Ignore the random trailing commas, I don't use trailing commas, but I copy and pasted parts of the shared config I use and VS Code insisted on putting in trailing commas.

```js
import typescriptESLint from "@typescript-eslint/eslint-plugin";
import typescriptParser from "@typescript-eslint/parser";
import eslintConfigPrettier from "eslint-config-prettier";
import prettier from "eslint-plugin-prettier";
import svelte from "eslint-plugin-svelte";
import globals from "globals";
import svelteParser from "svelte-eslint-parser";

/** @type {import("eslint").Linter.FlatConfig[]} */
export default [
  {
    ignores: globalIgnores,
    languageOptions: {
      ecmaVersion: "latest",
    },
    linterOptions: {
      reportUnusedDisableDirectives: true,
    },
  },
  {
    ignores: globalIgnores,
    languageOptions: {
      ecmaVersion: "latest",
    },
    plugins: { prettier },
    rules: {
      ...prettier.configs.recommended.rules,
      "prettier/prettier": "error",
    },
  },
  {
    files: svelteFiles,
    ignores: svelteIgnores,
    languageOptions: {
      ecmaVersion: "latest",
[...]

PR #588, opened 2023-10-07
fix: `parserPath` type might be `undefined` on Eslint Flat Config

krist7599555:
Fix this issue #584

to compatibility with Eslint Flat Config.
`parserPath` might be `undefined` and will throw error on worker. so check parserPath is string type before usage.

Maybe we can remove whole `svelte` block since [`eslint-plugin-svelte3`](https://github.com/sveltejs/eslint-plugin-svelte3) is deprecated

### DIFF
--- a/.changeset/thick-birds-share.md
+++ b/.changeset/thick-birds-share.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-prettier": patch
+---
+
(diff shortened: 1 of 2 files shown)


JounQin:
@krist7599555 Thanks for raising PR to fix the issue, but CI is broken unexpectedly, can you fix it?

krist7599555:
@JounQin might need a workflow approval

PR #592, opened 2023-10-18
feat: support flat config system

### LINKED ISSUE #591: Support ESLint flat config
Hey!

Thanks for a useful plugin. Right now, only eslintrc format is supported. In ESLint v9, the new flat format will be the default: https://eslint.org/blog/2023/10/flat-config-rollout-plans/

More and more projects are migrating to flat config format as a preparation. Do you have plans for supporting flat config in this plugin?

I'm willing to help out if this is something you want to do.

filiptammergard:
Resolves #591.

Wanted to open this for review to get some early feedback. This seems to work fine, if I create an `eslint.config.js` in the root with this content:

```js
const prettier = require("./flat");

module.exports = [
  prettier.configs.recommended,
];
```

I get ESLint errors accordingly.

The idea here is that the flat config file could be imported this way in other projects:

```js
const prettier = require("eslint-plugin-prettier/flat");

module.exports = [
  prettier.configs.recommended,
];
```

The reason not to change the main export is that the eslintrc format needs to have continued support. Eslintrc will be deprecated in v9 of ESLint and removed in v10, so to introduce a minimal amount of breaking changes, the flat config support can be purely additive and when ESLint v10 is released, the flat config could be made the main export in a new major version of this plugin.

Thoughts? If you think it's a good approach, I can continue adding documentation add changeset etc.


filiptammergard:
Any early input in this one @JounQin?

JounQin:
Sorry for forgetting this, I'll review recently ASAP.

pumano:
Would love to use it as it will be available! @filiptammergard thank you!

recommended.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const prettier = require('./eslint-plugin-prettier');
+// eslint-disable-next-line n/no-unpublished-require
```
JounQin, line 4: It seems we're missing it from `peerDependencies`

package.json
```diff
@@ -19,6 +19,7 @@
   "files": [
     "eslint-plugin-prettier.d.ts",
     "eslint-plugin-prettier.js",
+    "recommended.js",
```
JounQin, line 22: Can you add a `recommended.d.ts` accordingly?

recommended.js
filiptammergard: Yeah I think this is an error from before, it's included in `peerDependenciesMeta` but not in `peerDependencies` as it should. I can add it – but I suspect it can be seen as a breaking change to introduce a peer dependency, no?

filiptammergard: Added it here: https://github.com/prettier/eslint-plugin-prettier/pull/592/commits/35363bbe11a4cbed97ad679af778203a24d8c3a1

JounQin: It's a bugfix IMO, and we've declared it as optional already.

package.json
filiptammergard: Sure! How is `recommended.d.ts` used when it comes to an eslint plugin?

JounQin: https://github.com/eslint-types/eslint-define-config

recommended.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const prettier = require('./eslint-plugin-prettier');
+const prettierConfig = require('eslint-config-prettier');
+
+module.exports = {
```
JounQin, line 6: https://github.com/prettier/eslint-plugin-prettier/blob/17c3c109cb9022301e804e418cae38ff489f8e69/eslint-plugin-prettier.js#L79


Please use `// @ts-check` and jsdoc for type checking to align with `.d.ts`

JounQin (changes requested):

filiptammergard:
Some new docs on how to migrate to flat config format just came out: https://eslint.org/docs/latest/extend/plugin-migration-flat-config. Did some updates to align with those recommendations.

recommended.js
filiptammergard: I'll give it a try!

recommended.js
```diff
@@ -3,14 +3,23 @@
...
+  configs: {},
   rules: {
-    'prettier/prettier': 'error',
-    'arrow-body-style': 'off',
-    'prefer-arrow-callback': 'off',
+    ...prettier.rules,
   },
 };
+
+Object.assign(plugin.configs, {
```
JounQin, line 13: Avoid mutations.

JounQin (changes requested):

recommended.js
filiptammergard: I'd generally agree, but this is explicitly what ESLint is recommending in the case where you need to reference back to a plugin from a plugin config: https://eslint.org/docs/latest/extend/plugin-migration-flat-config#migrating-configs-for-flat-config

JounQin: Well, that's not always correct or suitable for everyone.

JounQin (changes requested):

recommended.js
JounQin: Well, that's not always correct or suitable for everyone

pumano:
@filiptammergard don't stop in your idea to support it)

recommended.js
BPScott: It feels like now might be a good time to get out of the business of shipping a recommended config that tries to pull in `eslint-config-prettier`.

It should be possible to use `eslint-plugin-prettier` without also needing `eslint-config-prettier` (because you might configure eslint to disable all the conflicting rules some other way).

But the move to flat config makes it so that if we keep the recommended config we have to go and add a peerDependency on `eslint-config-prettier`, rather than relying on the previously kinda weird implicit loading stuff.


@JounQin what would you think to the idea of us removing the recommended config, and instead we document how people should install and enable `eslint-config-prettier` themselves if they want that ruleset.

BPScott (commented):
Per https://eslint.org/blog/2023/10/flat-config-rollout-plans/

> You should also make sure that your rules aren’t using context.parserOptions and context.parserPath. Instead, you should be using context.languageOptions and context.languageOptions.parser, which also work when ESLint is run in eslintrc mode. See our [previous post](https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/) for more information.

Looking at https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/, it seems we need to update our usage of the now-deprecated:

- `context.getSourceCode`
- `context.getFilename`
- `context.getPhysicalFilename`

We will need to identify when the replacements for these functions were added, and update our minimum supported version of eslint accordingly. 

Any change to the minimum supported version will be a breaking change (Which is fine, but we'll need to update the changeset).

recommended.js
JounQin: I'm fine with the suggestion.

filiptammergard:
> Per https://eslint.org/blog/2023/10/flat-config-rollout-plans/
> 
> > You should also make sure that your rules aren’t using context.parserOptions and context.parserPath. Instead, you should be using context.languageOptions and context.languageOptions.parser, which also work when ESLint is run in eslintrc mode. See our [previous post](https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/) for more information.
> 
> Looking at https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/, it seems we need to update our usage of the now-deprecated:
> 
> * `context.getSourceCode`
> * `context.getFilename`
> * `context.getPhysicalFilename`
> 
> We will need to identify when the replacements for these functions were added, and update our minimum supported version of eslint accordingly.
> 
> Any change to the minimum supported version will be a breaking change (Which is fine, but we'll need to update the changeset).

Good input @BPScott!

Looks like the methods you're mentioning are becoming properties (added in v8.40.0): https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/#context-methods-becoming-properties

The methods will be removed in ESLint v10 (not v9), so there's no rush here. The blog post also suggest supporting both versions like this:

```js
const sourceCode = context.sourceCode ?? context.getSourceCode();
```

If the property is undefined (as per before ESLint v8.40.0), the function is used instead, and no breaking change needs to be considered.

However, I think those changes can be considered separately from supporting flat config format, so I made a separate PR for it: https://github.com/prettier/eslint-plugin-prettier/pull/594

recommended.js
filiptammergard: That sounds really good – I found the implicit use of `eslint-config-prettier` without properly defining it as a peer dependency to be strange.

What do you think about leaving `eslint-config-prettier` out for the flat config while keeping the eslintrc config as is for now, to be able to release flat config support in a minor version update?

Then you can decide wether you'd like to removing implicit use of `eslint-config-prettier` in eslintrc config, or if you want to remove the eslintrc config altogether when it's deprecated in ESLint v9.

recommended.js
filiptammergard: I removed it from peer dependencies here: https://github.com/prettier/eslint-plugin-prettier/pull/592/commits/e90c10892614d5ed71395123f1a112a71d8760dc

filiptammergard:
@JounQin @BPScott Updated now to avoid mutations and to align with the recommendation to always provide meta information: https://eslint.org/docs/latest/extend/plugin-migration-flat-config#adding-plugin-meta-information I needed to bump `@types/eslint` to get support for flat config types.

Also renamed from `recommended` to `flat` since that makes it more clear that this in a flat config export.

package.json
```diff
@@ -43,9 +44,6 @@
   "peerDependenciesMeta": {
     "@types/eslint": {
       "optional": true
-    },
-    "eslint-config-prettier": {
```
JounQin, line 47: It's actually required by legacy recommended config at

https://github.com/prettier/eslint-plugin-prettier/blob/17c3c109cb9022301e804e418cae38ff489f8e69/eslint-plugin-prettier.js#L82

JounQin (changes requested):

pnpm-lock.yaml
```diff
@@ -1,7 +1,7 @@
 lockfileVersion: '6.0'
 
 settings:
-  autoInstallPeers: false
+  autoInstallPeers: true
```
JounQin, line 4: Why this setting is changed?

flat.js
```diff
@@ -0,0 +1,35 @@
...
+ * @typedef {import('eslint').ESLint.Plugin} Plugin
+ */
+
+const prettier = require('./eslint-plugin-prettier');
+
+/**
+ * @type {Plugin}
+ */
+const plugin = {
+  meta: {
```
JounQin, line 14: This can be added into core file

flat.js
```diff
@@ -0,0 +1,35 @@
...
+    name: 'eslint-plugin-prettier',
+    version: '5.0.1',
```
JounQin, line 16: read from `package.json` instead

filiptammergard: Done

flat.js
filiptammergard: Done

pnpm-lock.yaml
filiptammergard: Was due to my pnpm settings, fixed now

package.json
filiptammergard: I can add it back if you want, but specifying a peer dependency as optional in `peerDependenciesMeta` doesn't actually do anything unless also specified in `peerDependencies`: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta

Should I take it back anyway?

filiptammergard: Sure, I'll add it to both `peerDependencies` and `peerDependenciesMeta` for now then. What version of `eslint-config-prettier` should be required from peer dependencies?

JounQin: I believe '*' is fine enough? How do you think?

filiptammergard: Yep, probably fine. Added it now!

package.json
```diff
@@ -19,6 +19,7 @@
   "files": [
     "eslint-plugin-prettier.d.ts",
     "eslint-plugin-prettier.js",
+    "flat.js",
```
JounQin, line 22: Please add `flat.d.ts`

.changeset/kind-apples-itch.md
```diff
@@ -0,0 +1,21 @@
...
+
+module.exports = [prettier.configs.recommended];
+```
+
+Or with ESM:
+
+```js
+import prettier from 'eslint-plugin-prettier/flat';
+
+export default [prettier.configs.recommended];
```
JounQin, line 20: It's still strange to me that only `prettier.configs.recommended` is used but the whole `prettier` plugin object is exported, however the plugin inside is actually the legacy one.

JounQin (changes requested):

BPScott (changes requested):
Thanks for fighting through this!


https://eslint.org/docs/latest/extend/plugin-migration-flat-config#backwards-compatibility makes it seem like it would be totally fine to have both a regular and flat config defined within the configs key of `eslint-plugin-prettier.js` instead of needing a whole new file for this.


It seems like it should be possible to add a new "recommended-flat" key to the configs object and define the config there and use it like so, instead of needing to add a new file to the repo:


```js
import eslintPluginPrettier from 'eslint-plugin-prettier';

export default [
  ...eslintPluginPrettier.configs.['recommended-legacy']
]
```

The linked blogpost suggested having a flat config called "BLAH" and the legacy equivalent is named named "BLAH-legacy". That seems like a nice idea eventually, but I like the idea of of avoiding a backwards compatibility break now by adding a new config that is flat-compatible and leaving the current one as-is. We can swap the names around in some future major version.


Could you please add the new flat compatible config to to the configs key in `eslint-plugin-prettier.js` with the name"recommended-flat", instead of adding a new entrypoint to the package.

recommended.js
BPScott: > What do you think about leaving eslint-config-prettier out for the flat config while keeping the eslintrc config as is for now, to be able to release flat config support in a minor version update?

Yes. I think that sounds ideal for this PR. Let's do that.

filiptammergard:
@BPScott I think both ways work, and for example `eslint-plugin-react` went the other way: https://github.com/jsx-eslint/eslint-plugin-react#configuring-shared-settings

But since ESLint recommends doing as you said, I think it might be a good idea to lean on that. The downside with having them in the same file is that I can't reference the plugin from the config like this:

```js
const eslintPluginPrettier = {
  configs: {
    recommended: {
      extends: ['prettier'],
      plugins: ['prettier'],
      rules: {
        'prettier/prettier': 'error',
        'arrow-body-style': 'off',
        'prefer-arrow-callback': 'off',
      },
    },
    'recommended-flat': {
      plugins: {
        prettier: eslintPluginPrettier,
      },
      rules: {
        'prettier/prettier': 'error',
        'arrow-body-style': 'off',
        'prefer-arrow-callback': 'off',
      },
    },
  },
  rules: {
    prettier: {/* ... */},
  },
};
```

That makes `eslintPluginPrettier` reference itself and that gives a circular dependency. So we have to either mutate it with `Object.assign`, or we create a new variable as I did now.

Thoughts on this?

filiptammergard:
@BPScott I saw you pushed a commit for using `Object.assign` instead of creating a new variable. Any other changes requested, or should I get going updating the docs accordingly?

pumano:
@JounQin could you please review it? Very important feature

eslint-plugin-prettier.js
```diff
@@ -224,4 +216,30 @@ const eslintPluginPrettier = {
...
+  recommended: {
+    extends: 'prettier',
+    plugins: ['prettier'],
+    rules: {
+      'prettier/prettier': 'error',
+      'arrow-body-style': 'off',
+      'prefer-arrow-callback': 'off',
+    },
+  },
+  ['recommended-flat']: {
```
JounQin, line 231: Should we use camelCase here?

eslint-plugin-prettier.js
```diff
@@ -224,4 +216,30 @@ const eslintPluginPrettier = {
   },
 };
 
+// Assign configs after initial plugin configuration, as flat configs
+// require referencing the original plugin that contains the rules.
+Object.assign(eslintPluginPrettier.configs, {
```
JounQin, line 221: How about `eslintPluginPrettier.configs = ` instead?

eslint-plugin-prettier.js
pumano: looks like config names should be as rules with dashes also eslint have config naming convention for eslint plugins with dashes too

JounQin: The previous rules are for legacy configs, for flat configs, we're using js object reference instead, so camelCase is preferred IMO.

filiptammergard: `recommended-flat` is all based on direction from the ESLint team:

> Export both eslintrc and flat configs. The configs key is only validated when a config is used, so you can provide both formats of configs in the configs key. We recommend that you append older format configs with -legacy to make it clear that these configs will not be supported in the future. For example, if your primary config is called recommended and is in flat config format, then you can also have a config named recommended-legacy that is the eslintrc config format.

https://eslint.org/docs/latest/extend/plugin-migration-flat-config#backwards-compatibility

They're recommending appending `-legacy` for legacy configs, which means `-flat` is the equivalent. Should we not go for that?

eslint-plugin-prettier.js
filiptammergard: Yeah that should be fine now since the original `eslintPluginPrettier.configs` object is empty. Updated it now!

eslint-plugin-prettier.js
JounQin: That strategy is an unnecessary breaking change IMO, and it can be made in next major version.

eslint-plugin-prettier.js
```diff
@@ -73,21 +74,12 @@ function reportDifference(context, difference) {
...
-      plugins: ['prettier'],
-      rules: {
-        'prettier/prettier': 'error',
-        'arrow-body-style': 'off',
-        'prefer-arrow-callback': 'off',
-      },
-    },
-  },
+  meta: { name, version },
+  configs: {},
```
JounQin, line 82:
```suggestion
```

eslint-plugin-prettier.js
```diff
@@ -73,21 +74,12 @@ function reportDifference(context, difference) {
 //  Module Definition
 // ------------------------------------------------------------------------------
 
-/**
- * @type {Plugin}
- */
+// Type hint configs key as non-nullable to avoid type-checking errors in when
+// assigning to eslintPluginPrettier.config below
+/** @type {Plugin & {configs: NonNullable<Plugin['configs']>}} */
```
JounQin, line 79:
```suggestion
/**
 * Type hint `configs` key as non-nullable to avoid type-checking errors in when
 * assigning to `eslintPluginPrettier.configs` below
 *
 * @type {Plugin & {configs: NonNullable<Plugin['configs']>}}
 */
```

Avoid mixed comments

eslint-plugin-prettier.js
filiptammergard: Yes, that's why the suggestion is to add `recommended-flat` now (non-breaking) and swap them around in next major.

JounQin: My suggestion is to add `flatRecommended` config key instead of `recommended-flat`.

eslint-plugin-prettier.js
filiptammergard: Fixed

BPScott:
Ah sorry, I thought I added a comment after pushing up that change at the weekend.

I've been thinking about this for the last few days, and now I've had a chance to play with what the eslint suggested approach looks like, I actually kinda hate it. Having to do object reassignment in `eslint-plugin-prettier` feels clunky, and the naming clashes of 'recommended' vs 'recommended-flat`/`flat-recommended` is awkward. 

This PR's original approach of mimicking eslint-plugin-react and creating a new `recommended.js` file, and getting people to import from `eslint-plugin-prettier/recommended` feels a bunch cleaner. We should feed this back to the eslint team at some point.


@JounQin I think I'm going back on my previous request of "lets follow eslint recommendations", and instead am preferring having the "recommended.js" file that was proposed at the start of this PR. Now you've saw both approaches, do you have on preference on which one feels better?

JounQin:
In my view I'd like to have a `flat.js` exporting a flat plugin, usage:

```js
const prettier = require("eslint-plugin-prettier/flat");

module.exports = [
  {
    // usage 1, use plugin manually
    plugins: {
      prettier,
    },
  },
  // usage 2, enable rules automatically
  prettier.configs.recommended,
];
```

BPScott:
I think "flat plugin" is a bit of a misnomer. If we add the `meta` block to the plugin then the existing plugin works with flat or legacy configs - at that point it's only the "configs" that we need to deal with.  I'm pretty sure that for usage one, people can already do the following (once we add the meta block), there's no need for a dedicated file:

```js
const prettier = require("eslint-plugin-prettier");

module.exports = [
  {
    // usage 1, use plugin manually
    plugins: {
      prettier,
    },
  },
];

```


And then usage 2 can be


```js
const prettierRecommended = require("eslint-plugin-prettier/recommended");

module.exports = [
  {
    ...prettierRecommended
  },
];
```

JounQin:
@BPScott Fair enough, I'm fine with that approach.

.changeset/kind-apples-itch.md
```diff
@@ -0,0 +1,21 @@
...
+
+module.exports = [prettier.configs['recommended-flat']];
+```
+
+Or with ESM:
+
+```js
+import prettier from 'eslint-plugin-prettier';
+
+export default [prettier.configs["recommended-flat"];
```
uiolee, line 20:
```suggestion
export default [prettier.configs["recommended-flat"]];
```

filiptammergard: Thanks!

filiptammergard:
> Now you've saw both approaches, do you have on preference on which one feels better?

I personally try to always lean on conventions and recommendation, so I think deviating from ESLint's recommendations should only be done if it's really necessary.

> I've been thinking about this for the last few days, and now I've had a chance to play with what the eslint suggested approach looks like, I actually kinda hate it. Having to do object reassignment in eslint-plugin-prettier feels clunky, and the naming clashes of recommended vs recommended-flat/flat-recommended is awkward.

I see what you mean. However, I think the most important thing is to make it the least clunky from a user's point of view. If the convention is to expose a new flat config `recommended-flat` from the same path, then most plugins will do that, user's will be used to that, and hence it will be the path of the least resistance for the users. User's don't care about wether or not an object reassignment is happening internally.

> This PR's original approach of mimicking eslint-plugin-react and creating a new recommended.js file, and getting people to import from eslint-plugin-prettier/recommended feels a bunch cleaner.

I think `flat` is better than `recommended` in this case. The currently exposed config is already a "recommended config" and we need to make it clear that the difference between the current config and the new one is that the new one is in flat config format.

This would convey that: 

```js
const prettier = require("eslint-plugin-prettier/flat");

module.exports = [
  prettier.configs.recommended,
];
```

While this wouldn't:

```js
const prettier = require("eslint-plugin-prettier/recommended");

module.exports = [
  prettier.configs.recommended,
];
```

I just asked about this in the ESLint community discord. Let's see if I get an answer.

uiolee:
I would prefer the structure likes eslint-plugin-react, It is easier to be extend or remove.

eslint-plugin-prettier.js
```diff
@@ -239,4 +234,30 @@ const eslintPluginPrettier = {
...
+      'prettier/prettier': 'error',
+      'arrow-body-style': 'off',
+      'prefer-arrow-callback': 'off',
+    },
+  },
+  ['recommended-flat']: {
+    // In the flat config, don't extend from the eslint-config-prettier ruleset.
+    // The consumer should deal with extending from eslint-config-prettier themselves.
+    plugins: {
+      prettier: eslintPluginPrettier,
```
uiolee, line 253: would it be a circular ref?

filiptammergard:
> I would prefer the structure likes eslint-plugin-react, It is easier to be extend or remove.

In what way is it easier to extend and remove?

eslint-plugin-prettier.js
filiptammergard: Hmm, you're actually right! I'll make a comment about this below.

filiptammergard:
As @uiolee pointed out in https://github.com/prettier/eslint-plugin-prettier/pull/592#discussion_r1383675969, doing re-assignment like this creates a circular ref:

```js
const eslintPluginPrettier = {
  configs: {},
};

eslintPluginPrettier.configs = {
  ['recommended-flat']: {
    plugins: {
      prettier: eslintPluginPrettier,
    },
  },
};
```

It's effectively as doing this:

```js
const foo = { bar: {} }
foo.bar = { baz: foo }
```

Which makes `foo` infinitively deep.

Same goes for doing it with `Object.assign`.

Based on this, I suggest we definitely go for a separate file. Is that fine?

filiptammergard:
I also got this answer from ESLint team on wether one way of exposing plugin configs is fundamentally better than other options or not:

> It’s really up to what you prefer. With flat config, we aren’t really enforcing any conventions for configs in plugins, so you can set it up however you’d like. Just make sure you document it for your users.

Based on this, it seems like the suggestion is the doc is merely a suggestion – not necessarily a recommendation or convention. I this it's totally fine to expose a separate file instead.

I also asked wether or not the circular ref becomes a problem and got this answer:

> No. there is no traversal into plugins referenced in configs so we never hit any problems with circular references.

So it's most likely completely fine despite the circular ref, but IMO good to avoid when possible even though it doesn't necessarily break anything.

filiptammergard:
A suggestion is to bump `@types/eslint` separately, so that this PR can handle only what's explicitly related to introducing flat config format support. https://github.com/prettier/eslint-plugin-prettier/pull/602

Sounds good?

filiptammergard:
And same for this one: https://github.com/prettier/eslint-plugin-prettier/pull/603

filiptammergard:
I dropped the commit where `@types/eslint` was bumped from this PR now as it is already taken care of in https://github.com/prettier/eslint-plugin-prettier/pull/602.

.changeset/kind-apples-itch.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+feat: support ESLint flat config system
+
+`eslint-plugin-prettier` can be used in the flat config format like this with CJS:
+
+```js
+const prettier = require('eslint-plugin-prettier');
+
+module.exports = [prettier.configs['recommended-flat']];
```
JounQin, line 12:
```suggestion
module.exports = [prettier.config.flatRecommended];
```

I still recommend this config name.

JounQin (changes requested):

.changeset/kind-apples-itch.md
filiptammergard: But even better with a separate `flat.js` file, right?

JounQin: I don't have such preference, a clear document is enough IMO.

filiptammergard:
Waiting for input by @BPScott on what direction to take before I proceed. I'll vote for using a separate file `flat.js`.

filiptammergard:
Do you have any input @BPScott?

JounQin:
I think the deadline could be this weekend, otherwise I'll help to refactor and release in my view.

BPScott:
> Waiting for input by @BPScott on what direction to take before I proceed. I'll vote for using a separate file flat.js.

Aah sorry I totally missed this and the updates from the eslint discord. I was planning to come back to this PR last week but needing a break got to me.

---

@filiptammergard thank you for pushing on with this, and asking for opinions from the eslint discord. It's great to know that they're not pushing hard on the approach suggested in https://eslint.org/docs/latest/extend/plugin-migration-flat-config.

Doing a little audit of other popular plugins:

- [`eslint-plugin-react`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/configs/recommended.js), [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import/blob/main/config/recommended.js), [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/configs/recommended.js) have a `config(s)/recommended.js` file for the flat recommended config.
- -[`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest/pull/1245/files#diff-ee7ac2e490d03c854fe269207dff9dd802bace92ee45f82870bda24315d98075) and  [`eslint-plugin-jsx-a11y`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/891/files#diff-4ecc7109dee61828a3bbe5dee2f6c6407751d6680cbd4889f76d31c73cd3476a) don't yet support flat config but have PRs that have the same `config/recommended.js` approach.
- [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc?tab=readme-ov-file#flat-config) and [eslint-plugin-n](https://github.com/eslint-community/eslint-plugin-n/blob/446e7441ce85031761648a45100025f4e3d79b0c/lib/index.js#L68) Adds a `flat/recommended` config to the main export
- `typescript-eslint` has a [draft PR](https://github.com/typescript-eslint/typescript-eslint/pull/7935) that goes down a third approach of creating a new wrapper package

With three popular packages diverging from the recommendations I don't feel bad about diverging. We have a single rule so we won't ever need multiple configs (ignoring the legacy vs flat distinction) so I think placing that in a configs folder is overkill. The root of the repo is fine and good.


I think I still prefer creating a new separate file and usage being `const prettierRecommended = require('eslint-plugin-prettier/recommended');`. I think that's a little cleaner than `const prettierRecommended = require('eslint-plugin-prettier').configs['flat/recommended']`.

I would prefer that new file be named `recommended.js` instead of `flat.js`. Consider fast forwarding to two years time when ESLint v10 is released and the legacy config format is dead -  having a file called `flat.js` feels strange in that world - I'd expect that the filename should be the name of the config - `recommended` feels like it makes more sense to me.

---

My request is annoyingly "back to what we had originally before I opened my mouth":

- A `recommended.js` in the root of the repo
- Add a `recommended.d.ts` with types
- Add those two files to the package.json's files array
- We'll need to update the README to describe usage when using flat config

PR #594, opened 2023-10-26
chore: use context properties where context methods are deprecated

filiptammergard:
But keep supporting the methods for now.

I think this might be considered a `chore` as it won't affect users of the plugin until ESLint v10 is released.


JounQin (approved):

BPScott:
Generally yes and thanks! I'll test this with old and new eslint versions and merge if it all looks good over the weekend

BPScott (approved):
Thanks! I pushed up  a tweak to the comments, to avoid having long lines - I prefer comments living above code rather than at the end of a code line

filiptammergard:
Great @BPScott ⭐

PR #613, opened 2023-12-02
feat: introduce `fullControl` option, allowing fine-grain control from ESLint

### LINKED ISSUE #612: Allow formatting code snippets in markdown
**What version of `eslint` are you using?**

`8.55.0`

**What version of `prettier` are you using?**

`3.1.0`

**What version of `eslint-plugin-prettier` are you using?**

`5.0.1`

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**

https://stackblitz.com/edit/node-mhr6sb?file=README.md,eslint.config.js

**What source code are you linting?**

See the Stackblitz link above

With fine-grained `files` option in the flat config, Prettier should be able to format virtual files created by `eslint-plugin-markdown` to format the code snippet in Markdown.

**What actually happened?**

There is a hard-coded logic to ignore any virtual file created in markdown as well as a few other extensions: 

https://github.com/prettier/eslint-plugin-prettier/blob/1882a36d91b7b6372bf593f008902028c42fd53d/worker.js#L129-L152

I consider it a valid default as it might cause nested formatting when users enable `eslint-plugin-prettier` to all files.

But with the new Flat Config, you are able to only enable certain rules with certain options for specific types of files. In the case where we don't config Prettier to format the parent markdown, treating the virtual files as normal formattable code should be allowed.

**Proposed solutions**

I imagine we could introduce a flag to opt-out of this hard-coded logic, and give the control of ignoring a file to users.

I made a simple implantation, adding a new config called `fullControl`: https://github.com/prettier/eslint-plugin-prettier/commit/a3336077a81a416af6ef052f4030c32ef329eb28

Let me know if you have a better idea/naming. Thanks!

antfu:
A solution to fix #612

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -36,7 +36,7 @@ const { INSERT, DELETE, REPLACE } = generateDifferences;
 
 // Lazily-loaded Prettier.
 /**
- * @type {(source: string, options: Options, fileInfoOptions: FileInfoOptions) => string}
+ * @type {(source: string, options: Options, fileInfoOptions: FileInfoOptions, fullControl: boolean) => string}
  */
 let prettierFormat;
 
@@ -115,6 +115,7 @@ const eslintPluginPrettier = {
                 properties: {},
                 additionalProperties: true,
               },
+              fullControl: { type: 'boolean' },
             },
             additionalProperties: true,
           },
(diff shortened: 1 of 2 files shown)


JounQin:
Let discuss at #612 first.

PR #615, opened 2023-12-03
Use flat config for the project's eslint configuration

BPScott:
Convert the project's eslint config to the flat style.

This shall let us dogfood the recommended eslint config when we add it.

This is a 1:1 migration, with the exception of adding `.github/ISSUE_TEMPLATE.md` to the ignores. I'm not sure what I've managed to do here, but it seems we weren't linting that before but now we are.


JounQin:
@BPScott `.md` files are linted by ESLint through `remark-lint` integration. And `eslint-plugin-mdx` supports flat config at the same time.

package.json
```diff
@@ -69,7 +70,7 @@
     "eslint-mdx": "^2.1.0",
     "eslint-plugin-eslint-plugin": "^5.1.0",
     "eslint-plugin-mdx": "^2.1.0",
-    "eslint-plugin-n": "^16.0.1",
+    "eslint-plugin-n": "^16.3.1",
```
JounQin, line 73: The CI is broken after upgrading.

JounQin (changes requested):
Please add mdx plugin back.

package.json
BPScott: [eslint-plugin-n v16 dropped support for node <16](https://github.com/eslint-community/eslint-plugin-n/releases/tag/16.0.0), and 16.1.0 introduced some syntax - `??=` that is only available in v15 and up. Annoyingly v16.1.0 is also the version that introduced the `flat/recommended` config that we need so we can't pin to a version that's got just the thing we need.

BPScott:
Gotcha, I hadn't clocked that eslint-plugin-mdx also lints against the markdowny portions of the markdown file rather than just the codeblocks. I'll add it back.

JounQin (approved):

BPScott:
There we go, I believe that's the markdown stuff put back properly. That last rebase was tidying up the wording of the global ignores block at the bottom @JounQin I've tagged you for one last re-review.

eslint.config.js
```diff
@@ -0,0 +1,52 @@
...
+    files: ['**/*.mdx'],
+    rules: { 'prettier/prettier': ['error', { parser: 'mdx' }] },
+  },
+  // Global ignores
+  // If a config block only contains any ignores key, then the globs are
+  // considerd as globally ignored
+  {
+    ignores: [
+      'CHANGELOG.md',
+      '.github/ISSUE_TEMPLATE.md',
```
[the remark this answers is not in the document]
JounQin: dotfiles are ignored by ESLint by default but not desired?

eslint.config.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+  },
+  {
+    files: ['**/*.md'],
+    rules: { 'prettier/prettier': ['error', { parser: 'markdown' }] },
+  },
+  {
+    files: ['**/*.mdx'],
+    rules: { 'prettier/prettier': ['error', { parser: 'mdx' }] },
+  },
```
BPScott, line 41: Had to add these as [this part of the recommended config](https://github.com/mdx-js/eslint-mdx/blob/master/packages/eslint-plugin-mdx/src/configs/recommended.ts#L21-L48) wasn't ported to their flat configs

JounQin: They are actually unnecessary now.

eslint.config.js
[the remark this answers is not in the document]
BPScott: With the legacy config format eslint default ignores `node_modules` and all files/folders beginning with a dot.
With the new config they've changed that to just `node_modules` and `.git`.

The [RFC line detailing this change](https://github.com/eslint/rfcs/blob/main/designs/2019-config-simplification/README.md?plain=1#L534) suggests that it's deliberate.

eslint.config.js
JounQin: OK, I'll take a look later, it's unexpected. Let's keep it as-is for  not blocking this PR.

eslint.config.js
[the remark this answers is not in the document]
JounQin: Yeah, I mean why not enable linting this time?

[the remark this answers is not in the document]
JounQin: Sure, sounds good to me.

JounQin (approved):

Issue #617, opened 2023-12-04
Not working with PUG template


enrico-hic:
**What version of `eslint` are you using?**
8.55.0

**What version of `prettier` are you using?**
3.1.0

**What version of `eslint-plugin-prettier` are you using?**
5.0.1

**Please paste any applicable config files that you're using (e.g. `.prettierrc` or `.eslintrc` files)**
.prettierrc

```
{
  "tabWidth": 2,
  "printWidth": 100,
  "singleQuote": true,
  "semi": false,
  "trailingComma": "all",
  "bracketSpacing": true,
  "bracketSameLine": false,
  "arrowParens": "always",
  "plugins": ["@prettier/plugin-pug"],
  "pugSortAttributes": "asc"
}
```

.eslintrc.json

```
{
  "env": {
    "browser": true,
    "commonjs": true,
    "es2021": true
  },
  "extends": [
    "eslint:recommended",
    "plugin:prettier/recommended"
  ],
  "plugins": [
    "pug"
  ],
  "parserOptions": {
    "ecmaVersion": "latest"
  }
}
```

**What source code are you linting?**
The plugin is working incorrectly in the files that use template pug (jade).
In the other files (like js) it is working correctly.

**What did you expect to happen?**
Report errors using the command `eslint file.pug` and highlight errors in the editor (VS Code).

**What actually happened?**
The errors are not highlighted in the editor but when I save the file is automatically formatted correctly using prettier.
The command eslint doesn't return any errors:

```
> eslint file.pug

npm timing command:run Completed in 202ms
npm verb exit 0
npm timing npm Completed in 223ms
npm info ok
```

But the command prettier report some errors:
```
> prettier file.pug --check

Checking formatting... : timing npm:load Completed in 18ms
[warn] file.pug
[warn] Code style issues found in the above file. Run Prettier to fix.
npm timing command:run Completed in 182ms
npm verb exit 1
npm timing npm Completed in 203ms
npm verb code 1
```

**Steps to reproduce**
You can find an example at this link: https://codesandbox.io/p/devbox/quizzical-joana-zlpzn8

Javascript working as expected:
1. Run `npm run prettier:js` -> errors are found
2. Run `npm run lint:js` -> errors from prettier are reported

Pug not working:
1. Run `npm run prettier:pug` -> errors are found
2. Run `npm run lint:js` -> no errors reported

enrico-hic:
[...]

PR #619, opened 2023-12-19
chore: skip formatting inline scripts in pug files

JounQin:
related: https://github.com/valpackett/eslint-plugin-pug and #617


BPScott (approved):

BPScott:
Could we add some tests, in the same style as the eslint-plugin-svelte tests?

JounQin:
> Could you add a test, in the same style as the eslint-plugin-svelte test?

OK, I'll add a test case when I'm free.

JounQin:
@BPScott I just added the test case for pug files, do you want to review again?

BPScott (approved):
Thanks for adding the test case!

PR #623, opened 2023-12-21
fix: add package.json exports field

### LINKED ISSUE #622: use es6's import syntax to import eslint-plugin-prettier/recommended
I hope to use es6's import syntax to import `eslint-plugin-prettier/recommended` instead of require syntax

BPScott:
Fix #622.

Add exports mapping to package.json, to allow `import eslintPluginRecommended from 'eslint-plugin-prettier/recommended'` to work as expected.

Strictly speaking this is a breaking change as it removes the ability for people to import from "eslint-plugin-prettier/eslint-plugin-prettier.js" and "eslint-plugin-prettier/recommended.js" but the former was never a good idea in the first place and the latter has only been available for a few days, so I'm not worried about closing off those two import paths in a non-major release.

### DIFF
--- a/.changeset/slimy-boats-fold.md
+++ b/.changeset/slimy-boats-fold.md
@@ -1,0 +1,7 @@
+---
+'eslint-plugin-prettier': patch
+---
+
+Add exports mapping to package.json, to allow `import eslintPluginRecommended from 'eslint-plugin-prettier/recommended'` to work as expected.
+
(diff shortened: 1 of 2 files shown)


JounQin:
Should `/recommend.js` be allowed at the same time?

BPScott:
> Should /recommended.js be allowed at the same time?

I don't think it should be. The only valid import paths for this package should be `eslint-plugin-prettier` and `eslint-plugin-prettier/recommended`

JounQin:
OK, I'm fine with this.

PR #629, opened 2023-12-24
chore: add `package.json` into `exports` map

JounQin:
It's very common to read the `package.json`, this is not a required but nice-to-have entry.


JounQin:
I'd merge it as-is, because it's already a user requested feature at #635.

BPScott:
Now we've got a user asking for it +1, thanks!

PR #635, opened 2024-01-10
export `package.json`

quisido:
I have a use case wherein I'm reading `eslint-plugin-prettier/package.json`'s dependency versions as a part of a unit test.

Currently, this results in the error `Missing "./package.json" specifier in "eslint-plugin-prettier" package` since `package.json` is not exported. This PR addresses this, making it possible to read the package metadata.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
       "types": "./eslint-plugin-prettier.d.ts",
       "default": "./eslint-plugin-prettier.js"
     },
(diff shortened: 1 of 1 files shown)


JounQin:
Duplicate of #629

JounQin:
https://github.com/prettier/eslint-plugin-prettier/releases/tag/v5.1.3

PR #664, opened 2024-06-11
Master

serconfindev01:


BPScott:
This appears to

- Rewrite the readme in spanish
- Delete contributing instructions
- Add a bunch of undesired vscode config
- Make eslint-plugin-prettier.js be esm instead of commonjs - which breaks support for the legacy config format used before eslint 9
- Introduce invalid and unparsable code to eslint.config.js.

Basically it breaks everything.

PR #689, opened 2024-10-18
Support 'config' Option

colinta:
Allows users to assign a custom path for the prettier config, ie a shared config in a monorepo.

Our use case is that we have a shared prettier config, but it doesn't live at the repository root.

Prettier supports this on the CLI using `--config` (and `--ignorePath`). This PR brings that feature to the eslint plugin.

e.g.
```bash
prettier --ignore-path config/prettier/prettierignore --config config/prettier/prettier.config.js --write src
```

Update: added a changelog w/ minor version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.3.0
+
+### Minor Changes
+
+- [#689](https://github.com/prettier/eslint-plugin-prettier/pull/689) [`acbcc76`](https://github.com/prettier/eslint-plugin-prettier/commit/acbcc76) From [@colinta](https://github.com/colinta)! - fix: support 'config' Option
+
 ## 5.2.1
 
 ### Patch Changes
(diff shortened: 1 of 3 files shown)


colinta:
I was using this locally (`yarn add --dev https://github.com/colinta/eslint-plugin-prettier`) and I'm seeing this error:

```
Oops! Something went wrong! :(

ESLint: 8.57.0

TypeError: prettier.resolveConfig is not a function
```

I was able to fix it by patching the import in worker.js:

```
    if (!prettier) {
      const imported = await import("prettier");
      prettier = imported.default;
    }
```

But this change is unrelated to the config that I added. 🤔

BPScott:
I'm generally against adding new options to eslint-plugin-prettier as placing config files in non-standard locations is basically looking for trouble - its really hard to coordinate properly across CLIs and editor settings and people often end up blaming the wrong thing. As such, I want to avoid creating these non-blessed paths to stop people from thinking it might be a good idea.

@colinta is a fellow Shopify person. I've chatted to him through internal channels and we've came up with solution to his needs that avoids adding this option, and have confirmed that I'm good to close this PR.

PR #692, opened 2024-11-04
chore: add name to recommended config preset

ST-DDT:
This PR adds a name to the eslint config preset.

The name will be shown in when inspecting the eslint config and for debugging purposes.

You can view the eslint preview using the following command:

````sh
npx @eslint/config-inspector@latest
````

<details>
<summary>UI-Diff</summary>

| Before | After |
| --- | --- |
| grafik | {4B409241-20BA-411B-86C4-352989EA3FF0} |

The remaining anonymous sections are from other eslint-plugin presets that don't have a name yet.

</details>

I also added the config group names to the eslint config itself for consistency.
I can revert that part if you don't like it.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -11,6 +11,7 @@ module.exports = [
   // eslint-community/eslint-comments does not expose a reusable flat config,
   // so create one from its legacy config
   {
+    name: '@eslint-community/eslint-comments',
     plugins: {
       '@eslint-community/eslint-comments': eslintPluginEslintComments,
     },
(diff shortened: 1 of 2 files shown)


ST-DDT:
> Personally I don't think we need that part.

Should I revert that part or should I wait for more opinions?

ST-DDT:
Is there anything I can do to fix these CI errors?

BPScott:
Thanks for this. I've dealt with this in #703.

PR #709, opened 2025-03-03
chore: update ESLint to v9

fisker:


.remarkrc
```diff
@@ -1,5 +1,2 @@
 {
-  "plugins": [
-    "@1stg/preset"
```
fisker, line 3: @JounQin This cause `lint` script fails, maybe dependencies is outdated? Would you mind take a look? https://github.com/prettier/eslint-plugin-prettier/actions/runs/13671651574/job/38222988172#step:7:15

test/prettier.mjs
```diff
@@ -27,64 +32,63 @@ const rule = eslintPluginPrettier.rules.prettier;
...
-      {
-        files: ['*.{md,mdx}'],
-        extends: 'plugin:mdx/recommended',
-        settings: {
-          'mdx/code-block': true,
-        },
-      },
-      {
-        files: '**/eslint-plugin-svelte3/*.svelte',
-        plugins: ['svelte3'],
```
fisker, line 68: `eslint-plugin-svelte3` is deprecated https://github.com/sveltejs/eslint-plugin-svelte3?tab=readme-ov-file#%EF%B8%8F-please-use-eslint-plugin-svelte-instead

test/prettier.mjs
```diff
@@ -381,15 +391,39 @@ runFixture(
...
+      },
+      line: 2,
+      message: 'Delete `;;;;;`',
+      messageId: 'delete',
+      nodeType: null,
+      ruleId: 'prettier/prettier',
+      severity: 2,
+    },
+  ],
+]);
```
fisker, line 412: I didn't see logic skipping the scripts, maybe the comment is incorrect?

fisker:
Need update [Branch protection rule](https://github.com/prettier/eslint-plugin-prettier/settings/branch_protection_rules/45324267) when merge.

JounQin:
Seems only housekeeping, but Node 14 compatible tests are disabled?

fisker:
> but Node 14/16 compatible tests are disabled?

ESLint can't run on them. Suggestion?

fisker:
> We may need to downgrade eslint to v8 for Node 14/16 on CI?

I can try, but do we have a plan to drop support? Since I didn't actually change the code, it should still work with ESLint 8, just not tested.

JounQin:
> I can try, but do we have a plan to drop support? Since I didn't actually change the code, it should still work with ESLint 8, just not tested.

Personally I'd follow `prettier`'s strategy instead of `eslint`'s.

What means when `prettier` changes its engines support, `eslint-plugin-prettier` will follow.

test/prettier.mjs
```diff
@@ -126,11 +72,12 @@ ruleTester.run('prettier', rule, {
       code: `('');\n`,
       filename: getPrettierRcJsFilename('single-quote', 'dummy.md'),
     },
-    // Should ignore files from node_modules
-    {
-      code: 'a();;;;;;\n',
-      filename: 'node_modules/dummy.js',
-    },
+    // TODO: enable this one https://github.com/eslint/eslint/issues/19471 get fixed
```
JounQin, line 75: Can we enable this for ESLint v8?

fisker: https://github.com/prettier/eslint-plugin-prettier/pull/709/commits/d1de552acf0d5fc2a2972390c620aa7b225904eb

JounQin (approved):

JounQin:
Thanks for your great work!

PR #710, opened 2025-03-06
chore: fix badge

fisker:
Let's also rename `master` to `main`? 

I can take care, leave a thumbup if you agree @JounQin @BPScott

- Update CI and settings.
- Update changeset config.
- Check related settings.

Should be all.


BPScott (approved):
+1 to renaming `master` to `main` for consistency with the other repos in the prettier org. It's been a tiny thing that I've never gotten around to. If you want to take it on that'd be great.

PR #714, opened 2025-03-22
chore: update ESLint to remove hack in tests

fisker:


JounQin (approved):

JounQin:
It seems ESLint v8 still requires that option, and current changes make the test seems too complex, so I'd like to keep it as-is.

But when we bump next major and drop ESLint v8 support, then we can remove it eventually.

PR #743, opened 2025-06-14
feat: support non-js languages

### LINKED ISSUE #312: Possible to process/lint CSS files?
Hi, I have a kind of weird use case for this eslint plugin.

[Eclipse Wild Web Developer](https://github.com/eclipse/wildwebdeveloper) supports ESLint for the Typescript language server (using [this](https://github.com/Quramy/typescript-eslint-language-service)). However, Wild Web Developer is missing [CSS formatting](https://github.com/eclipse/wildwebdeveloper/issues/38#issuecomment-650824449). 

I had the idea of using the prettier plugin for eslint to add CSS formatting warnings & possibly auto-CSS-format, however I'm not sure if it's possible to have eslint-plugin-prettier processing CSS files. 

Does anyone have an idea as to whether this is possible?

dotcarmen:
close #312 

This PR aims to support non-JS languages being linted with the `prettier/prettier` rule

Currently, this plugin only runs on `Program` nodes, which is typically the root node for languages that are processed into JS in ESLint. However, other eslint plugins may use a different root node type, such as `@eslint/json`'s `Document` node type

This PR expands support for other languages by making 2 changes:
1. instead of only matching `Program` nodes when instantiating the rule, the rule now matches the type of the root node that was parsed for the entire file.
2. instead of relying on `@eslint/js`'s `getLocFromIndex`, use a new local function `getLocFromOffset` that transforms a byte offset into an API-compatible location.

The latter change was necessary because eg `@eslint/json` doesn't include a `getLocFromIndex` method on its `SourceCode` class.

----

> [!IMPORTANT]
> This PR expands the `prettier/prettier` ESLint rule to support non-JS languages by matching the root node type and introduces `getLocFromOffset` to handle byte offsets.
> 
>   - **Behavior**:
>     - Expands `prettier/prettier` rule to support non-JS languages by matching the root node type of the parsed file instead of only `Program` nodes.
>     - Replaces `getLocFromIndex` with `getLocFromOffset` to convert byte offsets to locations, supporting languages without `getLocFromIndex`.
>   - **Functions**:
>     - Adds `getLocFromOffset()` to calculate line and column from byte offset in `eslint-plugin-prettier.js`.
>     - Modifies `reportDifference()` to use `getLocFromOffset()` instead of `getLocFromIndex`.
>   - **Misc**:
>     - Updates rule instantiation to use `[sourceCode.ast.type]` instead of `Program` in `eslint-plugin-prettier.js`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=prettier%2Feslint-plugin-prettier&utm_source=github&utm_medium=referral)<sup> for ee09d116f6a4d1ff090c3306018ea2bc4a61f1eb. You can [customize](https://app.ellipsis.dev/prettier/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

- **New Features**
  - Added support for linting and formatting JSON files with the Prettier ESLint rule.
  - Enhanced compatibility with CSS files via the @eslint/css package.
- **Bug Fixes**
  - Improved compatibility with different ESLint API versions for location reporting.
- **Tests**
  - Introduced new test cases for JSON formatting, including valid, invalid, and empty JSON files.
- **Chores**
  - Added development dependencies on the @eslint/json and @eslint/css packages.


eslint-plugin-prettier.js
```diff
@@ -57,10 +67,27 @@ let prettierFormat;
...
+  const [lineStart, line] = lineIndexes.reduce((acc, lineStart, lineNo) => lineStart < offset ? [lineStart, lineNo] : acc, [-1, -1]);
+  if (line < 0) {
+    const line = Math.max(0, lineIndexes.length - 1);
+    const column = offset - Math.max(0, lineIndexes[line]);
+    return { line, column, offset }
+  }
+
+  const column = offset - lineStart;
+  return { line, column, offset };
+}
```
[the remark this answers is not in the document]
dotcarmen: lol i *just* caught this locally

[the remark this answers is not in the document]
dotcarmen: that code is still wrong though. here's the function body that works:
```ts
function getLocFromOffset(lineIndexes, offset) {
  lineIndexes = [-1, 0, ...lineIndexes];

  let line = 0;
  while (line + 1 < lineIndexes.length && lineIndexes[line + 1] < offset) {
    line++;
  }

  const column = offset - lineIndexes[line];
  return { line, column, offset };
}
```

eslint-plugin-prettier.js
```diff
@@ -12,6 +12,13 @@
  */
 
 /**
+ * @typedef {{
```
JounQin, line 15: I think one line code is enough for such tiny typedef.

eslint-plugin-prettier.js
```diff
@@ -71,8 +100,16 @@ function reportDifference(context, difference) {
...
+  if (!('text' in sourceCode)) {
+    throw new Error('prettier/prettier: non-textual source code is unsupported');
+  }
+
+  const lineIndexes = [...sourceCode.text.matchAll(/\n/g)].map(match => match.index);
+  // first line in the file starts at byte offset 0
+  lineIndexes.unshift(0);
   const [start, end] = range.map(index =>
-    (context.sourceCode ?? context.getSourceCode()).getLocFromIndex(index),
+    getLocFromOffset(lineIndexes, index)
```
JounQin, line 112: I think we should only callback when it's unavailable.

JounQin (changes requested):
Thanks for working on this first, please add related test cases.

eslint-plugin-prettier.js
dotcarmen: `start` and `end` are necessary for `report`, so we need to compute them somehow. Since the only reliably-implemented methods are for computing ranges of a node, and prettier errors may overlap or be between nodes, I believe I went with the simplest solution here.

Unless I'm misunderstanding you?

JounQin: I mean something like `sourceCode.getLocFromIndex?.(index) ?? getLocFromIndex(sourceCode, index)`

JounQin: `sourceCode.getLocFromIndex?.(index)` can also be included into `getLocFromIndex(sourceCode, index)` instead.

test/prettier.mjs
```diff
@@ -25,11 +25,14 @@ import eslintPluginPug from 'eslint-plugin-pug';
...
 import eslintMdx from 'eslint-mdx';
+import eslintPluginJson from '@eslint/json';
 
 const rule = eslintPluginPrettier.rules.prettier;
 const RuleTester =
   eslintUnsupportedApi.FlatRuleTester ?? eslintPackage.RuleTester;
-const ESLint = eslintUnsupportedApi.FlatESLint ?? eslintPackage.ESLint;
+const ESLint = eslintUnsupportedApi.FlatESLint ?? eslint.ESLint;
+
+const isESLint9 = !eslintUnsupportedApi.FlatRuleTester;
```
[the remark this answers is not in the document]
JounQin: What're taling about? It's alreay been fixed by `var eslint`, there is no such runtime error any longer.

JounQin:
@dotcarmen Thanks for your contribution!

PR #759, opened 2025-09-01
(PR): add config flat

luas10c:
Adds simplicity when importing eslint configurations into new eslint configuration.

- [x] add flat/recommended config

----

> [!IMPORTANT]
> Adds `flat/recommended` ESLint configuration to `eslint-plugin-prettier.js` with `prettier` plugin and specific rules.
> 
>   - **Configuration**:
>     - Adds `flat/recommended` config to `eslint-plugin-prettier.js`.
>     - Includes `prettier` plugin and rules: `'prettier/prettier': 'error'`, `'arrow-body-style': 'off'`, `'prefer-arrow-callback': 'off'`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=prettier%2Feslint-plugin-prettier&utm_source=github&utm_medium=referral)<sup> for 9b7b811435b7d1a8b1f27d6b22e737870a83b26a. You can [customize](https://app.ellipsis.dev/prettier/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a flat/recommended ESLint configuration to support ESLint’s flat config format.
  * Exposes an equivalent, dedicated config named "prettier/recommended" for easy adoption.
  * Retains existing rule behavior: enforces Prettier and keeps arrow-body-style and prefer-arrow-callback disabled.

### DIFF
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -141,6 +141,7 @@ const eslintPluginPrettier = {
       },
     },
     'flat/recommended': {
+      name: 'prettier/recommended',
       plugins: {
         get prettier() {
           return eslintPluginPrettier
--- a/eslint-plugin-prettier.js
+++ b/eslint-plugin-prettier.js
@@ -140,6 +140,18 @@ const eslintPluginPrettier = {
         'prefer-arrow-callback': 'off',
       },
     },
+    'flat/recommended': {
+      plugins: {
+        get prettier() {
+          return eslintPluginPrettier
+        }
+      },
+      rules: {
+        'prettier/prettier': 'error',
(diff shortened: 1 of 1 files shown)


luas10c:
> Flat config is already supported per https://github.com/prettier/eslint-plugin-prettier?tab=readme-ov-file#configuration-new-eslintconfigjs
> 
> I don't want to add two ways of doing the same thing. We might change our approach in the future, but considering this package has basically zero maintenance burden that feels like busywork and making breaking changes just for the sake of it.

The form that was added does not follow the conventional pattern followed by the rest of the other libs. I think it would make a lot of sense to follow the same pattern.

Issue #764, opened 2025-09-28
Support standard flat file format


peaklabs-dev:
It would be good if importing the config worked the same way as with all the other plugins and rules. It would be much cleaner.

**Standard way:**
```ts
import { defineConfig } from "eslint/config";
import js from '@eslint/js';
import tseslint from 'typescript-eslint';
import svelte from 'eslint-plugin-svelte';
import prettier from 'eslint-plugin-prettier';

export default defineConfig(
  js.configs.recommended,
  ...tseslint.configs.strict,
  ...tseslint.configs.stylistic,
  ...svelte.configs.recommended,
  ...svelte.configs.prettier,
  ...prettier.configs.recommended, // This does not work at the moment.
	{
//
	},
);
```

**Current way:**
```ts
import { defineConfig } from "eslint/config";
import js from '@eslint/js';
import tseslint from 'typescript-eslint';
import svelte from 'eslint-plugin-svelte';
import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';

export default defineConfig(
  js.configs.recommended,
  ...tseslint.configs.strict,
  ...tseslint.configs.stylistic,
  ...svelte.configs.recommended,
  ...svelte.configs.prettier,
  eslintPluginPrettierRecommended
	{
//
	},
);
```

Issue #758, opened 2025-09-01
(FR): Flat config support


luas10c:
What do you think of the support for flat config? Like other various eslint plugin packages have done?

```js

 {
    name: 'prettier',
    files: ['**/*.{js,ts}'],
    plugins: {
      prettier
    },
    languageOptions: {
      globals: {
        ...globals.node,
        ...globals.es2022
      }
    },
    rules: {
      ...prettier.configs.recommended.rules
    }
 }

```

Esse codigo acima ficaria assim.

```js
prettier.configs.recommended or prettier.configs['flat/recommended']
```

This code above would look like this.

Eslint has completely killed support for the old configuration. I don't think it makes sense to maintain support for it, there could be a breaking changes release.

PR #765, opened 2025-09-28
feat: support standard flat file format

peaklabs-dev:
## Changes

- feat: Adds support for standard flat file format.

## Issues

- fixes https://github.com/prettier/eslint-plugin-prettier/issues/764

### Notes

- This PR is similar to https://github.com/prettier/eslint-plugin-prettier/pull/759 but does not add a new rule, it simply extends the existing code to support the standard flat-file format as well.
- I tested both the current implementation and the standard implementation and they both still work.
- I understand your comment about not having 2 ways to do things but I think we should recommend this method moving forward as it is consistent with every other plugin. The old method can be deprecated at some point much later or never.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,17 +80,18 @@ This will:
 
 ## Configuration (new: `eslint.config.js`)
 
-For [flat configuration](https://eslint.org/docs/latest/use/configure/configuration-files-new), this plugin ships with an `eslint-plugin-prettier/recommended` config that sets up both `eslint-plugin-prettier` and [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) in one go.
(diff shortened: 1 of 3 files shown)


BPScott:
> I forgot that we support non-flat config files, which will be broken by this change. Perhaps this PR would be better suited for a major version. 

Indeed this is a breaking change.

See https://github.com/prettier/eslint-plugin-prettier/issues/758#issuecomment-3243183519 and https://github.com/prettier/eslint-plugin-prettier/pull/759#issuecomment-3243184586 for extra flat-config context.

At some point we will do a major version bump to drop support for ESlint v8, node 14 and 16 and remove the need for `@types/eslint`. However at this point in time this plugin is stable, it doesn't need any new features, and thus this legacy support isn't getting in the way. Once there is a new feature that requires these major updates we'll make them.


I'm going to close this, as I'll likely end up re-implementing it myself once that time for major version bumps comes.

PR #790, opened 2026-05-16
ci: declare workflow-level `contents: read` on 2 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 3 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

## Summary by CodeRabbit

* **Chores**
  * Improved workflow security by configuring GitHub Actions workflow-level token permissions to read-only access for repository contents across CI pipelines. This reduces the scope of automated workflow credentials, aligning with security best practices and minimizing the potential impact of compromised workflow tokens.

[Review Change Stack](https://app.coderabbit.ai/change-stack/prettier/eslint-plugin-prettier/pull/790?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   ci:
     name: Lint and Test on Node ${{ matrix.node }} (ESLint ${{ matrix.eslint }})
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -7,6 +7,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   publish:
     runs-on: ubuntu-latest


.github/workflows/autofix.yml
```diff
@@ -11,6 +11,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
```
JounQin: This workflow will update PR automatically, so I think `read` is not enough.

JounQin (changes requested):

arpitjain099:
Good catch @JounQin, you're right. Dropped the `autofix.yml` change from the commit (it pushes fixes back to the PR via autofix-ci so needs write scope), kept the change on the other two workflows. Force-pushed with the commit SSH-signed (now Verified). Let me know if there's anything else.

arpitjain099:
@JounQin gentle ping - I dropped the `autofix.yml` change back on 5/18 and rebased; the PR is now scoped to just `ci.yml` + `pkg-pr-new.yml`, both of which are read-only. The CHANGES_REQUESTED label is from before that fix. Let me know if you'd like any more changes or want to close.

JounQin (approved):

